PR #81, opened 2021-07-08
chore: show problematic behavior

metacosm:
To see the issue: build using `mvn clean install`. The extension itself builds just fine but when building an app that uses it the build won't work as shown by the `integration-tests` module failing to build properly.


metacosm:
- git co 979ce51; mvn clean install => integration-tests fail
- git co 27cf4ba; mvn clean install => everything should work again
- git co 3354177; mvn clean install => now things fail at runtime, the state from application.properties / annotations is not properly captured anymore which leads the extension to make wrong choices

PR #91, opened 2021-07-30
feat: make adding a main application optional

metacosm:
Since most operator applications don't really need to do anything in the
main method besides calling `start` on Operator, make it possible to not
provide a `QuarkusApplication` at all, thus removing more boilerplate.


geoand (approved):
LGTM!

deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
```diff
@@ -136,6 +137,12 @@ ConfigurationServiceBuildItem createConfigurationServiceAndOperator(
         liveReload.setContextObject(ContextStoredCRDInfos.class, storedCRDInfos); // record CRD generation info in context for future use
 
         additionalBeans.produce(AdditionalBeanBuildItem.unremovableOf(OperatorProducer.class));
+
+        // if the app doesn't provide a main class, add the StartupListener
+        if (index.getAllKnownImplementors(DotName.createSimple(QuarkusApplication.class.getName())).isEmpty()) {
+            additionalBeans.produce(AdditionalBeanBuildItem.unremovableOf(StartupListener.class));
```
geoand, line 143: Actually, you might want to add the scope here, which is likely `@Singleton`

metacosm: Better to add the scope on the class or add it when the bean is produced?

geoand: Since you are conditionally registering it as a bean, I propose doing it here

PR #100, opened 2021-08-18
fix: fixed typo

goldmann:


runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/StartupListener.java
```diff
@@ -24,7 +24,7 @@ public void onStartup(@Observes StartupEvent event) {
             final var branch = !version.getExtensionBranch().equals(Version.UNKNOWN)
                     ? " on branch: " + version.getExtensionBranch()
                     : "";
-            log.info("Quakus Java Operator SDK extension {} (commit: {}{})", version.getExtensionVersion(),
+            log.info("Quarkus Java Operator SDK extension {} (commit: {}{})", version.getExtensionVersion(),
```
metacosm, line 27: Quakus is the next great framework for duck programmers! 🤦🏼

metacosm (approved):

runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/StartupListener.java
goldmann: Oh, this is the story behind it... 😃

PR #157, opened 2021-11-18
Fix yaml

phillip-kruger:


metacosm (approved):
Thank you! /me hates YAML!

phillip-kruger:
Let me know if that worked. It should pull into the catalog now (I think you need a release)

PR #176, opened 2021-12-20
chore(ci): set up snapshot release

metacosm:


.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/setup-java@v2
+        with:
+          distribution: temurin
+          java-version: 11
+          cache: 'maven'
+
+      - name: Configure Git author
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
```
gsmet, line 34: I think you can drop this. You shouldn't need to commit anything with a simple deploy.

Otherwise it looks good.

metacosm: Thanks for the review!

Issue #175, opened 2021-12-18
Operator crashes at startup in native build


sclorng:
I'm migrating from 1.9.x to 2.0.x and when runing in native, it crashes at startup:

```
2021-12-18 13:40:13,153 INFO  traceId= [io.jav.ope.Operator] (main) Operator SDK 1.10.1 (commit: f7aceb4) built on Fri Dec 10 13:52:03 GMT 2021 starting...
2021-12-18 13:40:13,153 INFO  traceId= [io.jav.ope.Operator] (main) Client version: 5.10.1
2021-12-18 13:40:13,350 ERROR traceId= [io.jav.ope.Operator] (main) Error retrieving the server version: io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:314)
        at io.fabric8.kubernetes.client.utils.Serialization.unmarshal(Serialization.java:266)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:569)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:527)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:510)
        at io.fabric8.kubernetes.client.dsl.base.OperationSupport.restCall(OperationSupport.java:782)
        at io.fabric8.kubernetes.client.BaseClient.getVersionInfo(BaseClient.java:174)
        at io.fabric8.kubernetes.client.BaseKubernetesClient.getVersion(BaseKubernetesClient.java:494)
        at io.fabric8.openshift.client.DefaultOpenShiftClient.getVersion(DefaultOpenShiftClient.java:535)
        at io.javaoperatorsdk.operator.Operator.start(Operator.java:74)
        at foo.bar.QuarkusOperator.run(QuarkusOperator.java:29)
        at foo.bar.QuarkusOperator_ClientProxy.run(Unknown Source)
        at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:125)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:67)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:41)
        at foo.bar.QuarkusOperator.main(QuarkusOperator.java:23)
[...]

PR #178, opened 2021-12-24
Run it tests in native mode

scrocquesel:
This will allow to discover errors like https://github.com/quarkiverse/quarkus-operator-sdk/issues/175

I pushed first without the update to 2.6.1 that fix the VersionInfo serialization error. Build should failed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,5 +53,5 @@ jobs:
           key: maven-repo-${{ runner.os }}-${{ steps.get-date.outputs.date }}
 
       - name: Build with Maven
-        run: mvn -B formatter:validate verify --file pom.xml
+        run: mvn -B formatter:validate verify -Dnative --file pom.xml
 


scrocquesel:
There is an extension in the integration-tests that allows the VersionInfo to be unserialized. The sample is not working while the integration test passed :(

scrocquesel:
Adding OperatorSDKResource.java to the sample fix the issue. I have no clue as to why VersionInfo is added to reflection items in this case.
Should we investigate more on this now that the main branch has been updated ?

scrocquesel:
I guess that a new version should be released (quarkus, fabric.io, sdk are all aligned) and continue investigation latter.

.github/workflows/build.yml
```diff
@@ -53,5 +53,5 @@ jobs:
           key: maven-repo-${{ runner.os }}-${{ steps.get-date.outputs.date }}
 
       - name: Build with Maven
-        run: mvn -B formatter:validate verify --file pom.xml
+        run: mvn -B formatter:validate verify -Dnative --file pom.xml
```
metacosm, line 56: This should be done in a different job, so that we can run both native and jvm tests to make sure we cover everything

metacosm (changes requested):

scrocquesel:
To me, it does both already. Jvm build and test, then native compilation and native test.

________________________________

metacosm:
Ah yes, my bad. What else is needed for this PR, then?

scrocquesel:
Nothing.

Télécharger Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

metacosm (approved):

PR #228, opened 2022-02-09
feat: enable CodeQL analysis

metacosm:


csviri:
Do we need it also to JOSDK?

metacosm:
> Do we need it also to JOSDK?

No idea, just checking what it does here because I don't know much about it :)

PR #229, opened 2022-02-09
feat: make it possible to not automatically start the operator

### LINKED ISSUE #207: Allow to disable creation & start of the Operator on the CDI
On Test, we need to create a new operator, with a new client pointing to a new namespace everytime a test method is executed.

Currently we are not using the Operator created by the OperatorProducer and we are creating a new instance, but we are still having some reconcilers registered for Default namespace.

It would be nice if we can disable the Producing of the Operator bean or at least the automatic start on the startup event on CDI, using a test profile property

metacosm:
The configuration option is `quarkus.operator-sdk.start-operator=false`.
Fixes #207


jonathanvila (changes requested):
Thanks for the work !!!!

Code LGTM , although I'm missing any test that really checks it.

Issue #243, opened 2022-03-01
Rename the extension `quarkus-operator-sdk-csv-generator` to `quarkus-operator-sdk-bundle-generator`


Sgitario:
Follow-up action by https://github.com/quarkiverse/quarkus-operator-sdk/pull/238

Instead of generating only the CSV manifests at `target/manifests`, this extension should generate the bundle at `target/bundle` folder with the following content:
- `manifests`: same content as before at `target/manifests`
- `metadata/annotations.yaml`: with some default annotations and allow to add some more annotations via configuration. Example:
```yaml
annotations:
  operators.operatorframework.io.bundle.channel.default.v1: alpha
  operators.operatorframework.io.bundle.channels.v1: alpha
  operators.operatorframework.io.bundle.manifests.v1: manifests/
  operators.operatorframework.io.bundle.mediatype.v1: registry+v1
  operators.operatorframework.io.bundle.metadata.v1: metadata/
  operators.operatorframework.io.bundle.package.v1: joke-operator
```

- `bundle.Dockerfile`: to generate the bundle image. Content:
```
FROM scratch

LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
LABEL operators.operatorframework.io.bundle.package.v1=joke-operator
LABEL operators.operatorframework.io.bundle.channels.v1=alpha
LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha

COPY manifests /manifests/
COPY metadata /metadata/
```

With these changes, it would not be necessary to call `opm alpha bundle generate` any longer (document and workflow build action should be updated accordingly).

Issue #245, opened 2022-03-01
Add documentation about how to deploy operators using OLM


Sgitario:
Follow-up action by https://github.com/quarkiverse/quarkus-operator-sdk/pull/238

At the moment, this is documented only in [the Joke example](https://github.com/quarkiverse/quarkus-operator-sdk/tree/main/samples/joke), but this should be documented in the official documentation of the Quarkiverse extension.

Note that the documentation should clearly document that the extension `quarkus-operator-sdk-bundle-generator` should only be used in a release profile or similar, this is when users intend to perform a release of the operator.

PR #247, opened 2022-03-01
feat: replace csv generator by bundle generator

Sgitario:
Fix https://github.com/quarkiverse/quarkus-operator-sdk/issues/243


samples/mysql-schema/src/main/resources/application.properties
```diff
@@ -1,6 +1,7 @@
 quarkus.operator-sdk.crd.apply=true
-quarkus.operator-sdk.generate-csv=true
 quarkus.container-image.builder=jib
 quarkus.native.additional-build-args=--initialize-at-run-time=org.apache.commons.lang3.RandomStringUtils
 
 quarkus.datasource.username=root
+quarkus.operator-sdk.bundle.channels=alpha
```
metacosm, line 6: Are these required? Maybe there should be default values?

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -79,16 +81,17 @@ CSVMetadataBuildItem gatherCSVMetadata(CombinedIndexBuildItem combinedIndexBuild
...
+            BuildTimeOperatorConfiguration operatorConfiguration,
             OutputTargetBuildItem outputTarget,
             CSVMetadataBuildItem csvMetadata,
-            BuildProducer<GeneratedCSVBuildItem> doneGeneratingCSV,
+            BuildProducer<GeneratedBundleBuildItem> doneGeneratingCSV,
             GeneratedCRDInfoBuildItem generatedCustomResourcesDefinitions,
             List<GeneratedKubernetesResourceBuildItem> generatedKubernetesManifests,
             BuildProducer<GeneratedFileSystemResourceBuildItem> generatedCSVs) {
-        if (configuration.generateCSV.orElse(false)) {
+        if (configuration.enabled) {
```
metacosm, line 92: Wondering if the default for this shouldn't be `false` as we don't want to be generating the bundle for every build… Maybe we should only activate the extension during a `package` build?

metacosm:
@all-contributors add @Sgitario for code, ideas

samples/mysql-schema/src/main/resources/application.properties
Sgitario: This is a good question. Yes, the channels and package properties are required, but I'm not sure if providing default values is the right approach. But, as you said, let's put some defaults and let's see how users use it, so:
What about if we use the application name as package name by default?
For the channels, let's use just "alpha" by default.
Wdyt?

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
Sgitario: It would be odd to me that we add the bundle dependency and that nothing different happens :/
What do you mean by enabling this extension only a `package` build? 
A final note: generating the bundle does not mean that we're going to deploy the operator, but users can use the generated bundle manifests to proceed with the process.

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/BundleGenerationConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "operator-sdk.bundle", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class BundleGenerationConfiguration {
+    /**
+     * Whether the extension should generate the Operator bundle.
+     */
+    @ConfigItem(defaultValue = "true")
```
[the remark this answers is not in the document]
Sgitario: Let's continue the discussion in the original comment.

samples/mysql-schema/src/main/resources/application.properties
metacosm: That seems reasonable indeed.

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
metacosm: Yes, but I'm trying to keep the build as light as possible, i.e. it shouldn't do things that are not needed. For example, the CRD is only generated if the extension detects that the classes impacting it have changed, otherwise the CRD generation is skipped. Same goes for the controller configuration. 

I agree that disabling the extension by default would be weird but it also doesn't make sense to generate the bundle on every build. It really only makes sense to generate it whenever we ask for the executable to be generate i.e. during a `mvn package` build. So, ideally, the bundle extension would only activate during `mvn package`.

samples/mysql-schema/src/main/resources/application.properties
Sgitario: Updated

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
Sgitario: As discussed, we'll document the bundle extension to be used only when releasing operators. This will be documented in https://github.com/quarkiverse/quarkus-operator-sdk/issues/245

metacosm:
Actually, let's merge this as-is and do the other changes in subsequent PRs.

samples/joke/src/main/java/io/quarkiverse/operatorsdk/samples/joke/JokeRequestReconciler.java
```diff
@@ -28,7 +28,7 @@
 import io.javaoperatorsdk.operator.api.reconciler.ControllerConfiguration;
 import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
-import io.quarkiverse.operatorsdk.csv.runtime.CSVMetadata;
+import io.quarkiverse.operatorsdk.bundle.runtime.CSVMetadata;
```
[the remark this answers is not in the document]
metacosm: See #256.

PR #258, opened 2022-03-03
improvements

### LINKED ISSUE #257: Remove Apache commons dependency

metacosm:
- refactor: use java.util.Set operation instead of Sets from Arc
- refactor: use String.join instead of stream operations
- refactor: rename getYAMLData to more appropriate getManifestData
- refactor: remove unneeded field
- refactor: use append instead of concatenation
- fix: remove unused warnings
- fix: wrong field was used
- refactor: minor improvements
- refactor: remove dependency on commons-lang3 Fixes #257
- fix: logging

### DIFF
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
@@ -85,7 +85,7 @@ public class CsvManifestsBuilder extends ManifestsBuilder {
             }
         }
 
-        if (metadata.installModes == null || metadata.maintainers.length == 0) {
+        if (metadata.installModes == null || metadata.installModes.length == 0) {
             csvSpecBuilder.addNewInstallMode(true, DEFAULT_INSTALL_MODE);
         } else {
             for (CSVMetadataHolder.InstallMode installMode : metadata.installModes) {
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
@@ -17,7 +17,6 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
 
-import org.apache.commons.lang3.StringUtils;
 import org.jboss.logging.Logger;
 
 import io.fabric8.kubernetes.api.model.ServiceAccount;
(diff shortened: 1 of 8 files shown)


samples/mysql-schema/pom.xml
```diff
@@ -30,10 +30,6 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-container-image-jib</artifactId>
     </dependency>
-    <dependency>
```
scrocquesel, line 33: How can it compile when it is still used ?

https://github.com/quarkiverse/quarkus-operator-sdk/blob/1d7c65a9c70a251ed84f66e869ccbc2d3a0f46ca/samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/MySQLSchemaReconciler.java#L11

metacosm: Hmm, actually, I made the wrong change, I was planning on leaving the samples alone. 😅
That said the dependency is probably coming transitively…

scrocquesel: yes it i brought by io.quarkus:quarkus-core-deployment.
+1 for letting it in the sample :)

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
```diff
@@ -162,13 +161,10 @@ private void handlePermissions(List<ClusterRole> clusterRoles, List<RoleBinding>
         Map<String, List<PolicyRule>> customPermissionRules = new HashMap<>();
         if (metadata.permissionRules != null) {
             for (CSVMetadataHolder.PermissionRule permissionRule : metadata.permissionRules) {
-                String serviceAccountName = StringUtils.defaultIfEmpty(permissionRule.serviceAccountName,
+                String serviceAccountName = defaultIfEmpty(permissionRule.serviceAccountName,
```
Sgitario, line 164: Can this line be written in a single line now?

Sgitario (approved):

samples/mysql-schema/pom.xml
metacosm: By the way, I'm having issues with the MySQL test locally: it times out here. I'm wondering if there's an issue with Docker and testcontainers on M1 macs…

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
metacosm: Done.

samples/mysql-schema/pom.xml
scrocquesel: > By the way, I'm having issues with the MySQL test locally: it times out here. I'm wondering if there's an issue with Docker and testcontainers on M1 macs…

@metacosm On some environnement (especially when the docker daemon is not on the host but in a wrapper VM like with Docker Desktop for win/mac), exposed containers are not accessible on localhost and you need to set an env var `TESTCONTAINERS_HOST_OVERRIDE=host.docker.internal` with the correct hostname.

Issue #242, opened 2022-02-23
The property `quarkus.operator-sdk.crd.apply` is ignored


Sgitario:
## How to reproduce?
You can reproduce this issue by using the [samples/joke](https://github.com/quarkiverse/quarkus-operator-sdk/tree/main/samples/joke) example:
1. `cd samples/joke`
2. mvn clean install -Dquarkus.operator-sdk.crd.apply=true

The auto-generated JokeRequest CRD is not installed.

## Problem
I think that the problem is that this is done when creating an instance of the `Operator` (see https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/OperatorProducer.java#L37). Therefore, this won't ever be applied at build time, but only at runtime when injecting the `Operator`.

metacosm:
The property is not ignored, but, indeed, it only works for dev mode on purpose (because applying a CRD automatically to a cluster is probably not a good idea when you're not developing your operator). It doesn't do anything if the operator is not running. Applying the CRD without the operator running would actually require to spawn a separate process, which, granted, could be done during the build.

Sgitario:
Ok, so I would add a comment in the property to state that this is only intended to work on DEV mode.

metacosm:
Do you think it would make sense to convert this issue into a feature request to support application of the CRD at build time? If not, we should close this issue.

Sgitario:
> Do you think it would make sense to convert this issue into a feature request to support application of the CRD at build time? If not, we should close this issue.

From my point of view, the CRD should always be applied/updated when running the application on DEV mode, otherwise is not guaranteed that the CRD is up to date with what the CRD that has been just auto-generated. 

So, when users use this property to the Maven command, I'd expect that the CRD is always applied even when I'm not running the application on DEV mode. 

In summary, I'd be ok to convert this issue into a feature request.

metacosm:
> > Do you think it would make sense to convert this issue into a feature request to support application of the CRD at build time? If not, we should close this issue.
> 
> From my point of view, the CRD should always be applied/updated when running the application on DEV mode, otherwise is not guaranteed that the CRD is up to date with what the CRD that has been just auto-generated.

This is indeed a valid point. Maybe I should just convert the default to `true` and people who absolutely do not want this feature could turn it off.

[...]

PR #260, opened 2022-03-07
feat: apply CRD at built time if `quarkus.operator-sdk.crd.apply=true`

Sgitario:
When enabling `quarkus.operator-sdk.crd.apply=true` at build time, we will register the CRD in the Kubernetes cluster where user is logged in. 

Updated documentation and added a note saying that this will always be done when running the application on Dev mode.

Fix https://github.com/quarkiverse/quarkus-operator-sdk/issues/242


core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/CRDConfiguration.java
```diff
@@ -31,6 +31,7 @@
 
     /**
      * Whether the extension should automatically apply updated CRDs when they change.
+     * When running on DEV mode, the CRD changes will always be applied automatically.
```
metacosm, line 34: This isn't true, though, because the default value for `apply` is still `false`. In fact, perhaps we should remove this property altogether and make it an automatic behavior during dev mode?

Sgitario: I removed this code: https://github.com/quarkiverse/quarkus-operator-sdk/pull/260/files#diff-e6fe47bd3b62e0de50164805a7f94703596cb45bd7deae0e5c9f0f554930773cL54
Therefore, this property is no longer used when running in DEV mode and hence the CRDs will always be applied.
I double-checked that this is now working like this.

metacosm (approved):
LGTM

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/CRDConfiguration.java
metacosm: Indeed, I missed it looking at the PR from the web interface's diffs.

PR #267, opened 2022-03-10
test(mysql-schema): refactor with standalone dependent resource

scrocquesel:
First step to raise issue with DepedentResource integration in Quarkus.

I tried to use the annotation syntax but JOSDK  `DependentResourceManager` is not CDI aware, so this is only using dependent resource in standalone mode.

### DIFF
--- a/samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/dependent/SecretDependentResource.java
+++ b/samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/dependent/SecretDependentResource.java
@@ -1,0 +1,76 @@
+package io.quarkiverse.operatorsdk.samples.mysqlschema.dependent;
+
+import java.util.Base64;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.apache.commons.lang3.RandomStringUtils;
+
+import io.fabric8.kubernetes.api.model.Secret;
+import io.fabric8.kubernetes.api.model.SecretBuilder;
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.Creator;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.Matcher.Result;
+import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResource;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.PrimaryToSecondaryMapper;
+import io.quarkiverse.operatorsdk.samples.mysqlschema.MySQLSchema;
+
+@ApplicationScoped
+public class SecretDependentResource extends KubernetesDependentResource<Secret, MySQLSchema>
+        implements PrimaryToSecondaryMapper<MySQLSchema>, Creator<Secret, MySQLSchema> {
+
+    public static final String SECRET_FORMAT = "%s-secret";
+    public static final String USERNAME_FORMAT = "%s-user";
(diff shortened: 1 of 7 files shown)


samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/dependent/SecretDependentResource.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    protected Class<Secret> resourceType() {
+        java.lang.reflect.Type type = getClass().getGenericSuperclass();
+        if (type instanceof Class<?>) {
+            type = ((Class<?>) type).getGenericSuperclass();
+        }
+
+        return (Class<Secret>) ((java.lang.reflect.ParameterizedType) type).getActualTypeArguments()[0];
+    }
+}
```
scrocquesel, line 76: This is mandatory because actual implementation of JOSDK `Utils.getFirstTypeArgumentFromExtendedClass` is not aware of proxy class when using Bean.

metacosm:
I will take a look but note that we need to support the latest version of the dependent support first :)

scrocquesel:
> I will take a look but note that we need to support the latest version of the dependent support first :)

What do you mean by 'the latest version of the dependent support first' ?

scrocquesel:
> > > I will take a look but note that we need to support the latest version of the dependent support first :)
> > 
> > 
> > What do you mean by 'the latest version of the dependent support first' ?
> 
> We needed #271 to be merged so that the latest version of the SDK API was used.

Yep, that's what I was missing. By far out of my reach.
Should I push a version with the `@dependent` annotation even if not working due to CDI unsupported ?

samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/dependent/SecretDependentResource.java
metacosm: Why aren't you returning `Secret.class` directly? The reflection is only used in the SDK to provide a default (though flawed) implementation. If you're providing an implementation of that method, you might as well return the actual type that we want, which in this case is `Secret.class`.

scrocquesel: Oh, I was trying to find a generic solution to be included in the SDK. I don't feel very comfortable having quarkus user override the resourceType when it is already set in every generic parameter. Too bad java don't allow to `return R.class`.

metacosm:
> I tried to use the annotation syntax but JOSDK `DependentResourceManager` is not CDI aware, so this is only using dependent resource in standalone mode.

Actually, I'm not sure what you're trying to accomplish with the annotations and CDI. If you use the annotations, you shouldn't need to get a reference to the `DependentResource` implementations i.e. you shouldn't need to `@Inject` them because everything should be managed. Could you elaborate on what you need to do in this use case, please?

scrocquesel:
> > I tried to use the annotation syntax but JOSDK `DependentResourceManager` is not CDI aware, so this is only using dependent resource in standalone mode.
> 
> Actually, I'm not sure what you're trying to accomplish with the annotations and CDI. If you use the annotations, you shouldn't need to get a reference to the `DependentResource` implementations i.e. you shouldn't need to `@Inject` them because everything should be managed. Could you elaborate on what you need to do in this use case, please?

DependentResource can have injected dependencies themself, that's why declarative mode should support CDI.  https://github.com/quarkiverse/quarkus-operator-sdk/pull/281 should allow to move the injected dependent resources in the reconciler to annotations. I close this PR and create one on cdi-dependents branch.

PR #269, opened 2022-03-11
fix: use josdk 2.1.2-SNAPSHOT

csviri:
The goal here is to make the snapshot version available to test with quarkus extension


metacosm:
Merging this for now but we probably need to have a dedicated branch for this to make it automated.

PR #288, opened 2022-03-30
fix: wait until app is running and archive logs

Sgitario:


.github/workflows/build.yml
```diff
@@ -84,6 +84,9 @@ jobs:
           nohup mvn quarkus:dev > app.log 2>&1 &
           PID=$(echo $!)
 
+          # Wait until the app is up and running
+          tail -f app.log | sed '/Installed features/ q'
```
Sgitario, line 88: the tail command will also print the content of the Maven dev command and will exit once the "Installed features" pharse is there.

.github/workflows/build.yml
```diff
@@ -84,6 +84,9 @@ jobs:
           nohup mvn quarkus:dev > app.log 2>&1 &
           PID=$(echo $!)
 
+          # Wait until the app installed the jokerequest custom resource definition
+          $SCRIPTS/waitFor.sh crd $K8S_NAMESPACE NAME jokerequests.samples.javaoperatorsdk.io
```
metacosm, line 88: what's `NAME` here?

Sgitario: it's the column of the kubectl get crd command

metacosm: why is that needed?

Sgitario: if the crd is not installed yet, it will print "resource does not exist". When it does exist, it will print the columns.

metacosm: OK, I think I got it, it's a hack to make the script work… can you add a comment for explaining why it's done this way, please?

Sgitario: This is not a hack, but a text expectation. Having said that, in this case, it was not necessary. I've already updated the command.

metacosm (approved):

PR #293, opened 2022-04-05
fix: update for latest SDK changes

metacosm:


core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
```diff
@@ -460,15 +460,19 @@ private QuarkusControllerConfiguration createControllerConfiguration(
...
+                        cfg = new QuarkusKubernetesDependentResourceConfig(dependentNamespaces, labelSelector);
                     }
 
                     var nameField = dependentConfig.value("name");
                     final var name = Optional.ofNullable(nameField)
                             .map(AnnotationValue::asString)
                             .filter(Predicate.not(String::isBlank))
                             .orElse(DependentResource.defaultNameFor(dependentClass));
+                    final var spec = dependentResources.get(name);
+                    if (spec != null) {
```
Sgitario, line 472: Why this change is needed?

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
```diff
@@ -398,7 +398,7 @@ private QuarkusControllerConfiguration createControllerConfiguration(
                     getLabelSelector(controllerAnnotation),
                     Optional.ofNullable(specClassName),
                     Optional.ofNullable(statusClassName),
-                    dependentResources);
+                    dependentResources.values().stream().collect(Collectors.toUnmodifiableList()));
```
Sgitario, line 401: Why not use directly a list instead of creating a map in the `createDependentResources` method?

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/QuarkusKubernetesDependentResourceConfig.java
```diff
@@ -1,16 +1,18 @@
...
 public class QuarkusKubernetesDependentResourceConfig extends KubernetesDependentResourceConfig {
 
     @RecordableConstructor
-    public QuarkusKubernetesDependentResourceConfig(String[] namespaces, String labelSelector) {
+    public QuarkusKubernetesDependentResourceConfig(Set<String> namespaces, String labelSelector) {
         super(namespaces, labelSelector);
     }
 
-    public String[] getNamespaces() {
+    public Set<String> getNamespaces() {
```
Sgitario, line 15: I've just found that it was not working for me because I deleted these two getters. Can you add a comment saying that the getters are necessary for the recordable constructor?

metacosm: Yes, I have similar comments on other classes, I wish Quarkus would make this explicit… maybe with an annotation.

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
metacosm: So that it's easier to check if a dependent resource with the same name is already configured.

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
metacosm: It wasn't needed but it's similar to what the SDK does: it shouldn't be possible to add dependent resources with the same name several times.

Issue #273, opened 2022-03-15
Generate a single CSV CR for multiple controllers/crds


andreaTP:
In the Keycloak operator we have(at the moment) 2 controllers and 2 CRDs.
The old Go operator used to reference all the CRDs in the `spec.customresourcedefinitions.owned` field ( [e.g.]((https://github.com/k8s-operatorhub/community-operators/pull/757/files#diff-5145e4855a6359acb4f5eb9c81e68b56c5aaad6710e986a2dd80a485a15b7d2aR120-R146)) ).
I would like to reproduce the same behaviour as the current CSV generator creates(by default) a separate csv CR each Controller.

Trying to use the same name in `CSVMetadata` ( done [here](https://github.com/keycloak/keycloak/commit/e34a8c12e906a0a0eb54328627801d607204cbb1) ) result in a single csv CR overwritten by the latest controller.

cc. @vmuzikar

PR #294, opened 2022-04-05
Support bundle generation for multiple reconcilers

Sgitario:
Support a single bundle generation for multiple reconcilers. 

By default, the reconcilers that are not annotated with `@CSVMetadata` will be grouped into a single bundle altogether. 

If users need to customise this single bundle, they have to make use of the `SharedCSVMetadata` interface:

```java
@CSVMetadata(name = "pingpong-operator")
public class PingPongOperatorCSVMetadata implements SharedCSVMetadata {
}
```

If no implementation of `SharedCSVMetadata` is present, then the name of this single bundle is taken from the property `quarkus.operator-sdk.bundle.package-name` and, if not set, it will use the module name. 

Moreover, all these changes have been covered via a new sample called ping pong. This new sample uses two reconcilers_ the PingReconciler that updates the ping status to PROCESSED and creates a new pong resource, and another pong reconciles that simply updates the pong status to PROCESSED.

Fix https://github.com/quarkiverse/quarkus-operator-sdk/issues/273
Documentation will be updated as part of https://github.com/quarkiverse/quarkus-operator-sdk/issues/245


Sgitario:
cc @metacosm @andreaTP

andreaTP:
Hi @Sgitario ! This effort is really very much appreciated!
I will try to look into this soon enough!

metacosm (approved):

PR #307, opened 2022-04-08
fix: allow to generate multiple operators from different reconcilers

Sgitario:
Added unit test to verify this scenario.
Manually tested using the ping pong sample and annotating the two reconcilers with `@CSVMetadata(name = "ping-operator")` and the same with the pong reconciler.

### DIFF
--- a/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/First.java
+++ b/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/First.java
@@ -1,0 +1,17 @@
+package io.quarkiverse.operatorsdk.bundle.sources;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import io.fabric8.kubernetes.api.model.Namespaced;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Version;
+
+@Group("samples.javaoperatorsdk.io")
+@Version("v1alpha1")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class First extends CustomResource<Void, Void> implements Namespaced {
+
+    public First() {
+    }
+}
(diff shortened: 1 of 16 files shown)


bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/FirstReconciler.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkiverse.operatorsdk.bundle.sources;
+
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+import io.quarkiverse.operatorsdk.bundle.runtime.CSVMetadata;
+
+@CSVMetadata(name = "first-operator")
```
metacosm, line 8: I think the default test should actually be to check that we generate a CSV per reconciler in the absence of `CSVMetadata` annotations.

metacosm (changes requested):

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/FirstReconciler.java
Sgitario: added

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/First.java
```diff
@@ -0,0 +1,17 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import io.fabric8.kubernetes.api.model.Namespaced;
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Version;
+
+@Group("samples.javaoperatorsdk.io")
+@Version("v1alpha1")
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
metacosm, line 12: Why is this needed?

Sgitario:
duplicated by https://github.com/quarkiverse/quarkus-operator-sdk/pull/312

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/First.java
Sgitario: It's not needed. It's there because I copied it from the ping-pong example.

PR #312, opened 2022-04-12
fix bundle generator

metacosm:
- fix: allow to generate multiple operators from different reconcilers
- adding test to bundle extension to cover when csvmetadata is not used
- chore: check that all the files are indeed generated
- fix: proper file name for generated CRDs


Sgitario (approved):

metacosm:
There are at least a couple of possible issues with the current implementation but they should probably be addressed in subsequent PRs:

- #313 
- #314

PR #325, opened 2022-04-22
updates

metacosm:
- docs: add BOM information, note about Quarkus version
- chore: add test to check variable expansion in namespaces
- chore(deps): update to JOSDK 3.0.0.RC3 & Quarkus 2.8.1


metacosm:
The tests are currently failing because of https://github.com/java-operator-sdk/java-operator-sdk/issues/1179. Leaving the code as-is so that we can validate that the issue is indeed fixed in the next JOSDK release. See #326 for a workaround in the mean time. /cc @scrocquesel

PR #331, opened 2022-04-29
Fix for josdk 1179

scrocquesel:
In the reconciler, if the secret is not yet ready, an event should be fired when it will be the case.
The reconciler should not throw on purpose to retry but let new events the opportunity to reconcile

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,11 @@
   <version>4.0.0.Beta2-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Quarkus - Operator SDK - Parent</name>
+  <properties>
+    <quarkus.version>2.8.1.Final</quarkus.version>
+    <java-operator-sdk.version>3.0.0.RC3</java-operator-sdk.version>
+    <kubernetes-client.version>5.12.2</kubernetes-client.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -11,6 +11,11 @@
   <version>4.0.0.Beta2-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Quarkus - Operator SDK - Parent</name>
+  <properties>
+    <quarkus.version>2.8.2.Final</quarkus.version>
+    <java-operator-sdk.version>3.0.0.RC3</java-operator-sdk.version>
+    <kubernetes-client.version>5.12.2</kubernetes-client.version>
+  </properties>
```
metacosm, line 18: Not sure about this change… This make it easier to deal from our side but wouldn't that make it harder to consume the BOM from a user's perspective?

metacosm:
Also, can you please update that PR so that it targets the `updates` branch, instead? Thanks in advance!

pom.xml
scrocquesel: I confirm it still works

metacosm (approved):
LGTM

PR #345, opened 2022-05-17
feat: add initial WithDisposableNamespace test resource implementation

metacosm:


test/src/main/java/io/quarkiverse/operatorsdk/test/WithDisposableNamespace.java
```diff
@@ -0,0 +1,31 @@
...
+
+    int waitAtMostSecondsForDeletion() default 0;
+
+    String namespace() default UNSET_VALUE;
+
+    Fixture[] fixtures() default {};
+
+    int fixturesReadinessTimeoutSeconds() default 60;
+
+    @interface Fixture {
```
scrocquesel, line 26: Not sure it's possible because Java do not support polymorphism in annotation but would it be possible to do something near

```java
@WithDisposableNamespace(fixtures = {
        @YAMLResourceFixture(fromYAMLResource = "k8s/mysql-deployment.yaml"),
        @YAMLResourceFixture(fromYAMLResource = "k8s/mysql-service.yaml"),
        @PortFowardFixture(port = 3306, podSelector = @MatchLabel(key = "app", value ="mysql"), namespace = MY_SQL_NS)
})
```

maybe with some meta annotation ?

```java
@WithDisposableNamespace(fixtures = {
        @YAMLResourceFixture(fromYAMLResource = "k8s/mysql-deployment.yaml") factory = FromYamlResourceFactory.class,
        @YAMLResourceFixture(fromYAMLResource = "k8s/mysql-service.yaml") factory = FromYamlResourceFactory.class,
        @PortFowardFixture(port = 3306, podSelector = @MatchLabel(key = "app", value ="mysql"), namespace = MY_SQL_NS) factory = PortForwardFixtureFactory.class
})
```

If not, I understand that Fixture factories would be hard coded in some way ?

metacosm: Indeed, I don't think that this will be possible. However, I certainly could split the fixtures into several types: one for YAML resources, one for port forwarding and a programmatic one (that's what the factory is for: if present, it will load the resources by calling the `build` method on the specified class, basically to allow arbitrary code to be executed as part of the fixture setting for the tests). What do you think?

metacosm:
This PR is superseded by https://github.com/quarkusio/quarkus/pull/25938 since it makes sense to include that functionality in Quarkus directly.

PR #356, opened 2022-06-01
feat: Quarkus codestart support

metacosm:
- feat: add initial support for codestart
- refactor: move IT tests in the core module where they should be
- chore: remove unused classes
- feat: add initial code for generation
- chore(tests): add codestart test


metacosm:
> Looks good so far, except i saw a snapshot version for the codestart in the test. Something will have to be done about that, I suppose.

That's actually a pain… without specifying the version, this doesn't work. Ideally, we wouldn't need to specify the version and the codestart would automatically use the same version as the project.

aloubyansky:
To workaround it, in the meantime, you could pass a system property containing the project version from the surefire/failsafe config.

ia3andy:
This is already done but it will need to wait for 1.10:
https://github.com/quarkusio/quarkus/pull/25397

Until then:
https://github.com/quarkiverse/quarkus-quinoa/blob/main/integration-tests/pom.xml#L55
and
https://github.com/quarkiverse/quarkus-quinoa/blob/main/integration-tests/src/test/java/io/quarkiverse/quinoa/it/QuinoaCodestartIT.java#L17

metacosm:
This shouldn't be needed anymore since we can use CLI commands or Quarkus dev mode to create APIs once the project is bootstrapped.

PR #390, opened 2022-08-25
fix: keep the project compatible with Quarkus 2.7

rsynek:
@metacosm Quarkus 2.7, which is the latest productized version, brings an older version of Jandex. This PR makes the project compatible with Quarkus 2.7 at the expense of using a deprecated method (until a new productized version of Quarkus is released).

Maybe setting up a CI that overrides the Quarkus version and checks the compatibility would prevent similar issues in the future?

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/QuarkusControllerConfigurationBuilder.java
+++ b/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/QuarkusControllerConfigurationBuilder.java
@@ -284,7 +284,8 @@ class QuarkusControllerConfigurationBuilder {
             AnnotationConfigurableAugmentedClassInfo configurableInfo) {
         if (configurableInfo != null) {
             final var associatedConfigurationClass = configurableInfo.getAssociatedConfigurationClass();
-            if (reconcilerInfo.classInfo().annotationsMap().containsKey(associatedConfigurationClass)) {
+            // Keeping the deprecated method due to compatibility with Quarkus 2.7.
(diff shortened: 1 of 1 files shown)


metacosm:
@rsynek thank you for this! It's important to note, though, that the branch that is targeting 2.7 is the 3.0.x branch so no effort was made to assert compatibility with 2.7 in `main`. Of course, if we can maintain compatibility we will… I will explore the CI option indeed to see if we can at least get an idea of how compatible things are, indeed.

rsynek:
> @rsynek thank you for this! It's important to note, though, that the branch that is targeting 2.7 is the 3.0.x branch so no effort was made to assert compatibility with 2.7 in `main`. Of course, if we can maintain compatibility we will… I will explore the CI option indeed to see if we can at least get an idea of how compatible things are, indeed.

Thanks @metacosm , I think it's important to keep the main compatible with the latest Quarkus version, plus the equivalent of the latest productized version, if possible. Especially if we talk about a major version change of this project to have compatibility.

metacosm:
Is this change actually needed? From https://github.com/quarkiverse/quarkus-operator-sdk/pull/391, it appears that `main` builds just fine on 2.7.6…

metacosm:
> Is this change actually needed? From #391, it appears that `main` builds just fine on 2.7.6…

Actually, the CI setup was wrong (the Quarkus version wasn't actually changed 🤦🏼)

rsynek:
@metacosm would it be possible to release 4.0.1 with this fix included, please? Or even better, are there any plans for a regular release cadence?

metacosm:
@rsynek 4.0.1 is released. Thank you for your help!

PR #408, opened 2022-09-19
Compatibility with Jandex 3

Ladicek:


Ladicek:
For the record, this requires Jandex 2.4.3 and hence breaks compatibility with Quarkus 2.10 and before.

metacosm:
Thank you!

PR #413, opened 2022-09-25
feat: add dev services

scrocquesel:
Enable a k8s cluster (api server only, k3s, or kind) when a valid configuration is not found.

This use the k8s testcontainer provided by the amazing work done at https://github.com/dajudge/kindcontainer under the hood.

Support hot reload if the config do not change. Currently, the SDK do some optimisation and do not try to redeploy CRD if not changed. But if the config is changed and dev services needs to restart the container, the operator will not found the CRD anymore upon restart.

Support sharing, a second application in dev mode can reuse the same k8s cluster.

I have changed pingpong and mysqlschema to not use k8s mock server. MysqlSchema no longer needs to fake deletion of the CR.

### DIFF
--- a/core/deployment/pom.xml
+++ b/core/deployment/pom.xml
@@ -69,6 +69,15 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-devservices-deployment</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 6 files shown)


metacosm:
I guess this won't work with Quarkus 2.7…

scrocquesel:
I close it as you were faster than me aligning with the 2.15

metacosm:
Ah yes, sorry! Thank you so much for contributing that feature in Quarkus, @scrocquesel!

PR #417, opened 2022-09-27
fix: remove fabric8 k8s client version override

rsynek:
Removes the version override for `io.fabric8:kubernetes-*`. Versions of these dependencies are already contained in the `quarkus-bom`.

@metacosm This removes the version override we discussed.


metacosm:
Thank you, I've done it in the `next` branch but I guess this should also be done for `main`.

metacosm (approved):

PR #419, opened 2022-09-29
feat: inject leader election / informer stopped handler into config

metacosm:
Requires java-operator-sdk/java-operator-sdk#1493


vmuzikar (approved):
I played around with it with Keycloak Operator and can confirm it works well for our use-case. Thanks!

PR #423, opened 2022-10-06
Publish test-jars

aloubyansky:
@metacosm this will be enough to be integrated in the platform


metacosm:
Excellent. Thank you!

PR #429, opened 2022-10-19
CSVMetadata parameters required for operator publication

mgubaidullin:
ClusterServiceVersion requires additional parameters for publication. I have added them to CSVMetadata

### DIFF
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
@@ -198,6 +198,27 @@ public class BundleProcessor {
                     AnnotationValue::asString, () -> mh.providerURL);
         }
 
+        final var annotationsField = csvMetadata.value("annotations");
+        CSVMetadataHolder.Annotations annotations;
+        if (annotationsField != null) {
+            final var annotationsAnn = annotationsField.asNested();
+            annotations = new CSVMetadataHolder.Annotations(
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "containerImage",
+                            AnnotationValue::asString, () -> null),
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "repository",
+                            AnnotationValue::asString, () -> null),
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "capabilities",
+                            AnnotationValue::asString, () -> null),
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "categories",
+                            AnnotationValue::asString, () -> null),
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "certified",
+                            AnnotationValue::asBoolean, () -> false),
+                    ConfigurationUtils.annotationValueOrDefault(annotationsAnn, "almExamples",
+                            AnnotationValue::asString, () -> null));
+        } else {
+            annotations = mh.annotations;
+        }
+
         final var maintainersField = csvMetadata.value("maintainers");
         CSVMetadataHolder.Maintainer[] maintainers;
         if (maintainersField != null) {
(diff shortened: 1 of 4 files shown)


bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -214,6 +235,38 @@ private CSVMetadataHolder createMetadataHolder(AnnotationInstance csvMetadata, C
...
+                        ConfigurationUtils.annotationValueOrDefault(linkAnn[i], "name",
+                                AnnotationValue::asString, () -> null),
+                        ConfigurationUtils.annotationValueOrDefault(linkAnn[i], "url",
+                                AnnotationValue::asString, () -> null));
+            }
+        } else {
+            links = mh.links;
+        }
+
+        final var iconField = csvMetadata.value("icon");
```
metacosm, line 254: The icon is already dealt with automatically if users add an appropriately named file in `src/main/resources`. See https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java#L168-L180. Can we find a way to unify this with explicit metadata?

metacosm (changes requested):

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -214,6 +235,38 @@ private CSVMetadataHolder createMetadataHolder(AnnotationInstance csvMetadata, C
...
+        CSVMetadataHolder.Icon[] icon;
+        if (iconField != null) {
+            final var iconAnn = iconField.asNestedArray();
+            icon = new CSVMetadataHolder.Icon[iconAnn.length];
+            for (int i = 0; i < iconAnn.length; i++) {
+                icon[i] = new CSVMetadataHolder.Icon(
+                        ConfigurationUtils.annotationValueOrDefault(iconAnn[i], "base64data",
```
metacosm, line 261: In particular, passing base64 data in an annotation is not the most user-friendly thing out there ;)

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/CSVMetadataHolder.java
```diff
@@ -6,17 +6,50 @@ public class CSVMetadataHolder {
...
     public final String[] keywords;
     public final String providerName;
     public final String providerURL;
     public final String replaces;
     public final String version;
     public final String maturity;
+    public final String minKubeVersion;
     public final Maintainer[] maintainers;
+    public final Link[] links;
+    public final Icon[] icon;
```
metacosm, line 19: Seems like this fields would be better as a plural as well…

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/CSVMetadataHolder.java
```diff
@@ -67,28 +110,35 @@ public RequiredCRD(String kind, String name, String version) {
...
-        this(name, null, null, null, null, null, null, null, null, null, null, null, null);
+        this(name, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null);
     }
 
-    public CSVMetadataHolder(String name, String description, String displayName, String[] keywords, String providerName,
-            String providerURL, String replaces, String version, String maturity, Maintainer[] maintainers,
+    public CSVMetadataHolder(String name, String description, String displayName, Annotations annotations, String[] keywords,
+            String providerName,
+            String providerURL, String replaces, String version, String maturity, String minKubeVersion,
+            Maintainer[] maintainers, Link[] links, Icon[] icon,
```
metacosm, line 126: icon -> icons

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/CSVMetadataHolder.java
mgubaidullin: I intentionally used singular for `icon` (despite being an array) to reflect ClusterServiceVersion tag `icon` that is singular as well
Screen Shot 2022-10-21 at 9 11 49 AM

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
mgubaidullin: Perhaps we could provide an alternative options to user?
1. Select file from `src/main/resources`
Screen Shot 2022-10-21 at 9 28 41 AM

2. Use base64 string
Screen Shot 2022-10-21 at 9 29 26 AM

metacosm:
I've made some modifications to your latest changes but I cannot update your PR: can you make it so that I can update the code?
My changes are basically removing the `base64data` field on `Icon` because I don't think that users are likely to use that, not to mention the fact that we probably want to avoid adding that data in the app if we can avoid it. To that effect, I also changed the location of where the icons should be put in order for the generator to pick them up: it will now look for icons using the name defined in `Icon` in `src/main/kubernetes` instead of simply checking for a default icon in `src/main/resources`.

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/CSVMetadataHolder.java
mgubaidullin: Does 👍 mean you are agree and I could close it as resolved?

metacosm: Yes.

mgubaidullin:
Screen Shot 2022-10-21 at 4 38 54 PM

metacosm:
Not sure what's going on but I still can't push changes to your branch… 😭 
Will open a different PR with the additional changes and will close this one if you're OK with this.

metacosm:
See #432.

mgubaidullin:
Yes. I'm ok. Thanks

PR #445, opened 2022-11-23
fix: make native tests work in Quarkus ecosystem CI

metacosm:


integration-tests/pom.xml
```diff
@@ -91,6 +91,11 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
               <skipTests>${native.surefire.skip}</skipTests>
+              <environmentVariables>
+                <NAMESPACE_FROM_ENV>fromEnvVarNS</NAMESPACE_FROM_ENV>
+                <QUARKUS_OPERATOR_SDK_CONTROLLERS_EMPTY_NAMESPACES>fromEnv1,  fromEnv2</QUARKUS_OPERATOR_SDK_CONTROLLERS_EMPTY_NAMESPACES>
+                <VARIABLE_NS_ENV>variableNSFromEnv</VARIABLE_NS_ENV>
```
gsmet, line 97: Let's try that first and then you can also try to add them to the `systemPropertyVariables` block below.

integration-tests/pom.xml
```diff
@@ -91,6 +91,11 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
               <skipTests>${native.surefire.skip}</skipTests>
+              <environmentVariables>
+                <NAMESPACE_FROM_ENV>fromEnvVarNS</NAMESPACE_FROM_ENV>
```
gsmet, line 95: If you want to make sure both Surefire and Failsafe are consistent, you can probably do:
```suggestion
                <NAMESPACE_FROM_ENV>${env.NAMESPACE_FROM_ENV}</NAMESPACE_FROM_ENV>
```
But I would advise you to do it in a second step. Let's keep it simple first and iterate with simple stuff to understand how we can pass the values.

gsmet (approved):

PR #469, opened 2023-01-04
fix: retry should properly be reset when configuration changes

### LINKED ISSUE #467: max-retries is ignored in the controller configuration (operator-sdk version 4.0.5 and quarkus 2.13.6)
With the operator-sdk version 4.0.5 and quarkus 2.13.6, with the following controller 

```java

@ApplicationScoped
@ControllerConfiguration(name = FooController.NAME, labelSelector = "app.kubernetes.io/managed-by=foocontroller")
public class FooController implements Reconciler<Foo>,
                                      EventSourceInitializer<Foo>,
                                      ErrorStatusHandler<Foo> {

    public static final String NAME = "foocontroller";
    private static final Logger LOGGER = LoggerFactory.getLogger(FooController.class);

    @Override
    public UpdateControl<Foo> reconcile(Foo foo, Context<Foo> context) {
        LOGGER.info("Create or update Foo: '{}' in namespace '{}'",
                    foo.getMetadata().getName(),
                    foo.getMetadata().getNamespace());

        throw new RuntimeException("NICE TRY");
    }

```

and the following `application.properties`
```text
# Looks like this is ignored by quarkus...
quarkus.operator-sdk.controllers.foocontroller.retry.max-attempts=1
```

the reconcile loop is called 5 times instead of just one (when the exception is raised).

See https://github.com/r00ta/issues-reproducers/tree/main/operator-sdk-bug-ignore-max-retries for the reproducer

metacosm:
Fixes #467


r00ta:
Just tested with my reproducer: looks like it works properly with this fix :+1: Thank you!

PR #460, opened 2022-12-15
Implement a compile time check of the Quarkus version

andreaTP:
This is a naive implementation of a compile time check for the Quarkus version.

Unfortunately the version returned from `quarkus.platform.version` is `999-SNAPSHOT` and is not usable ...

PR #478, opened 2023-01-06
feat: build time check of versions alignment

metacosm:
- Implement a compile time check of the Quarkus version
- Fix dynamic version retrieval
- Conditional failure
- Commented out fabric8 check
- fmt


metacosm:
Supersedes #460.

Issue #477, opened 2023-01-06
Figure out if CsvManifestsBuilder.handleClusterPermission correct


metacosm:
The `groupToCRInfo` structure processing seems to have been removed during the work leading to #246. We should figure out whether this is still needed and if not, remove the associated code.

Sgitario:
To be precise, the `groupToCRInfo` was commented out in https://github.com/quarkiverse/quarkus-operator-sdk/pull/294 and then removed by a later pull request. 

Then, the remaining and unused `groupToCRInfo` map is only used in https://github.com/quarkiverse/quarkus-operator-sdk/blob/bd74e006dc02ee1adb236c27085ae047bc3ac156/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java#L345. 

Note that I'm quite unfamiliar with the logic of this method `CsvManifestsBuilder.handleClusterPermission`, but it was done to fix this issue https://github.com/quarkiverse/quarkus-operator-sdk/issues/145. I tried to reproduce this issue and it's working fine with the changes in the main, so I guess we can remove the associated code: https://github.com/quarkiverse/quarkus-operator-sdk/pull/479

metacosm:
This logic was used to deal with possible roles / bindings in a provided `kubernetes.yml` fragment, iirc, so I'm not sure we can actually remove it and the associated use case might not be working properly at the moment.

Sgitario:
If so, I doubt the logic ever worked as intended because it was using the rules from any cluster rule found in the provided `kubernetes.yml` (it should use only the cluster rule with a matching rule with the same service binding name). 
Even though after https://github.com/quarkiverse/quarkus-operator-sdk/pull/246, we can provide the rules directly in the CSVMetadata annotation, I think the use case you mention It's a nice one to support (code, test, and document), but I would recommend doing it from zero, and without using the `groupToCRInfo` structure.

PR #480, opened 2023-01-09
Fix having duplicate permissions in generated CSV

Sgitario:
Before these changes, the permissions were duplicated for clusterPermissions and permissions in the CSV manifests:

```yaml
clusterPermissions:
      - rules:
        - apiGroups:
          - apiextensions.k8s.io
          resources:
          - customresourcedefinitions
          verbs:
          - get
          - list
        serviceAccountName: quarkus-operator-sdk-samples-pingpong
       - rules:
        - apiGroups:
          - apiextensions.k8s.io
          resources:
          - customresourcedefinitions
          verbs:
          - get
          - list
        serviceAccountName: quarkus-operator-sdk-samples-pingpong
```

And should have only one:

```yaml
clusterPermissions:
      - rules:
        - apiGroups:
          - apiextensions.k8s.io
          resources:
          - customresourcedefinitions
          verbs:
          - get
          - list
        serviceAccountName: quarkus-operator-sdk-samples-pingpong
```

Moreover, the new logic will aggregate the rules onto the permission with the same service account name.

### DIFF
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import org.jboss.logging.Logger;
 
@@ -32,6 +33,8 @@ import io.fabric8.openshift.api.model.operatorhub.v1alpha1.ClusterServiceVersion
 import io.fabric8.openshift.api.model.operatorhub.v1alpha1.ClusterServiceVersionFluent;
 import io.fabric8.openshift.api.model.operatorhub.v1alpha1.ClusterServiceVersionSpecFluent;
 import io.fabric8.openshift.api.model.operatorhub.v1alpha1.NamedInstallStrategyFluent;
+import io.fabric8.openshift.api.model.operatorhub.v1alpha1.StrategyDeploymentPermissionsBuilder;
+import io.fabric8.openshift.api.model.operatorhub.v1alpha1.StrategyDeploymentPermissionsFluent;
 import io.quarkiverse.operatorsdk.bundle.runtime.CSVMetadataHolder;
 import io.quarkiverse.operatorsdk.bundle.runtime.CSVMetadataHolder.RequiredCRD;
 import io.quarkiverse.operatorsdk.common.ReconciledAugmentedClassInfo;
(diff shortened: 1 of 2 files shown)


Sgitario:
@metacosm I found this issue after investigating https://github.com/quarkiverse/quarkus-operator-sdk/issues/477

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
```diff
@@ -327,11 +329,16 @@ private void handlerPermission(List<PolicyRule> rules,
             String serviceAccountName,
             NamedInstallStrategyFluent.SpecNested<ClusterServiceVersionSpecFluent.InstallNested<ClusterServiceVersionFluent.SpecNested<ClusterServiceVersionBuilder>>> installSpec) {
         if (!rules.isEmpty()) {
-            installSpec
-                    .addNewPermission()
-                    .withServiceAccountName(serviceAccountName)
-                    .addAllToRules(rules)
-                    .endPermission();
+            if (installSpec.hasMatchingPermission(permissionHasServiceAccountName(serviceAccountName))) {
```
metacosm, line 332: Use a local variable for the result of `permissionHasServiceAccountName` to avoid creating it twice.

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
```diff
@@ -385,11 +392,16 @@ private void handleClusterPermission(List<PolicyRule> rules,
             }
         });
 
-        installSpec
-                .addNewClusterPermission()
-                .withServiceAccountName(serviceAccountName)
-                .addAllToRules(rules)
-                .endClusterPermission();
+        if (installSpec.hasMatchingClusterPermission(permissionHasServiceAccountName(serviceAccountName))) {
```
metacosm, line 395: Same as above.

metacosm (changes requested):
Looks like you're fixing the consequence (i.e. preventing outputting the permission twice) but what about the cause? Why were the permissions output twice in the first place, i.e. are we calling things too many times when not needed?

Sgitario:
> Looks like you're fixing the consequence (i.e. preventing outputting the permission twice) but what about the cause? Why were the permissions output twice in the first place, i.e. are we calling things too many times when not needed?

The cause is that the PingPong example has two reconcilers, so this logic is called twice which it's correct and expected.

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
Sgitario: Done

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/builders/CsvManifestsBuilder.java
Sgitario: Done

metacosm (approved):

metacosm:
Thank you!

PR #488, opened 2023-01-18
Option to enable parallel CRD generation

nicoloboschi:
* New option to enable the parallel crd generation introduced in https://github.com/fabric8io/kubernetes-client/pull/4644
* Added docs in the antora site (also for quarkus-operator-sdk_quarkus.operator-sdk.crd.generate-all that was missing)

This is draft since it requires the upgrade of kubernetes-client to 6.4


metacosm:
I need to check that Fabric8 PR again :)
Note that there are several options here: as the extension is driving the CRD generation, the parallelism could be introduced in the extension instead of at the CRD generator level, which wouldn't require changing the CRD code at all (at least, that I can think of at this point, but this obviously would need to be checked).

metacosm (approved):
LGTM

metacosm:
> @metacosm would you mind to review this again?
> I saw that the doc location has changed, I added the new config entry manually, was it the right way or it's generated automatically now?

It's automatically generated now. :)

nicoloboschi:
nice. I removed the doc part now

PR #625, opened 2023-06-15
Fix formatting

gsmet:


metacosm:
This has been addressed in #624. The issue was actually that CI wasn't actually building the proper branch but always building `main`. 🤦🏼

PR #650, opened 2023-07-10
fix: do not automatically stop Operator when it wasn't started automatically either

chendouble:
…tomatically configured


metacosm:
Good point. Thank for the catch!

PR #651, opened 2023-07-10
Add namespace configuration for generated manifests

Jamstah:
This enables the manifests generated at build time to be customised with namespaces to watch without affecting the controllers themselves. This way the controllers can use default behaviour allowing them to be configured at runtime, and the manifests can be used to apply that configuration at runtime.

Unfortunately, the logic to test for all namespaces and current namespaces in the Java operator SDK is in protected static methods, so I had to copy that logic into this project. It may be sensible to suggest that logic is made public in the Java operator SDK instead.

This came out of work being done for the keycloak oeprator here: https://github.com/keycloak/keycloak/pull/21231

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddRoleBindingsDecorator.java
+++ b/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddRoleBindingsDecorator.java
@@ -1,11 +1,15 @@
 package io.quarkiverse.operatorsdk.deployment;
 
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.DEFAULT_NAMESPACES_SET;
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.WATCH_CURRENT_NAMESPACE_SET;
 import static io.quarkiverse.operatorsdk.deployment.AddClusterRolesDecorator.getClusterRoleName;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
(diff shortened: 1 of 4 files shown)


metacosm:
Hi @Jamstah, unless I'm misunderstanding your issue, I think that the `quarkus.operator-sdk.controllers.<controller name>.generate-with-watched-namespaces` property is what you're looking for. See https://github.com/quarkiverse/quarkus-operator-sdk/releases/tag/6.2.1 (though this feature was available in 6.2.0 already).

Jamstah:
More likely to be my misunderstanding that yours :)

---

If I use this to build:
```
mvn clean package -Doperator -Dquarkus.container-image.build=true -Dquarkus.operator-sdk.manifest.generate-with-watched-namespaces=JOSDK_WATCH_CURRENT
```

And run with the env var:
```
QUARKUS_OPERATOR_SDK_NAMESPACES=""
```

I get this in the pod logs:
```
2023-07-10 20:11:26,012 INFO [io.jav.ope.Operator] (main) Registered reconciler: 'keycloakrealmimportcontroller' for resource: 'class org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImport' for namespace(s): [all namespaces]
2023-07-10 20:11:26,019 INFO [io.jav.ope.Operator] (main) Registered reconciler: 'keycloakcontroller' for resource: 'class org.keycloak.operator.crds.v2alpha1.deployment.Keycloak' for namespace(s): [all namespaces]
```

So at runtime, I successfully overrode the namespaces.

---

If I use this to build:
```
mvn clean package -Doperator -Dquarkus.container-image.build=true -Dquarkus.operator-sdk.controllers.keycloakcontroller.generate-with-watched-namespaces=jammy
```

And run with the env var:
```
QUARKUS_OPERATOR_SDK_NAMESPACES=""
```

I get this in the pod logs:
```
2023-07-10 20:20:27,219 INFO [io.jav.ope.Operator] (main) Registered reconciler: 'keycloakrealmimportcontroller' for resource: 'class org.keycloak.operator.crds.v2alpha1.realmimport.KeycloakRealmImport' for namespace(s): [all namespaces]
2023-07-10 20:20:27,253 INFO [io.jav.ope.Operator] (main) Registered reconciler: 'keycloakcontroller' for resource: 'class org.keycloak.operator.crds.v2alpha1.deployment.Keycloak' for namespace(s): [jammy]
```

So the namespace can't be override (and I don't think there's an option for watch current either)

---

In both cases, a RoleBinding not a ClusterRoleBinding is generated, which is what we're trying to find a property to do.

metacosm:
I'm still struggling to understand what you're trying to achieve outside of the different options you've tried. Could you explain, from a user's perspective, the use case you're trying to fulfil? Then we can figure out if something needs to change in QOSDK or if it's a matter of better documentation. Thanks in advance.

Jamstah:
Was writing that up in the keycloak PR, so to save duplicating the thread: https://github.com/keycloak/keycloak/pull/21231#issuecomment-1632423816

Jamstah:
Closed by #654

PR #658, opened 2023-07-14
Treat empty runtime namespaces as All Namespaces mode

### LINKED ISSUE #656: Operators do not work with OLM all namespaces mode
When using OLM, OLM will provide a pod annotation for the operator listing the namespaces the operator should watch. Typically, this is passed to the container as an environment variable by using the downward API.

OLM will put a comma separated list of namespaces into the annotation, reserving the empty string `""` as an indicator to mean "All namespaces on the cluster".

With a QOSDK operator, we use the `QUARKUS_OPERATOR_SDK_NAMESPACES` to configure the namespaces the operator should watch based on the annotation set by OLM, which will be `""` for all namespaces mode. This runtime configuration does not override any configuration set at build time, perhaps because the empty string is not picked up by quarkus, to be determined.

This means that the QOSDK does not work with OLM all namespaces mode when the operator is built with the `quarkus.operator-sdk.controllers.keycloakcontroller.generate-with-watched-namespaces` build time property, because the empty string does not override the setting correctly.

Jamstah:
This makes the empty string in the namespaces configuration watch all namespaces, as per the documented behaviour: https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java#L56

Fixes #656

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java
+++ b/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java
@@ -1,5 +1,8 @@
 package io.quarkiverse.operatorsdk.runtime;
 
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.DEFAULT_NAMESPACES_SET;
+import static io.quarkiverse.operatorsdk.runtime.Constants.QOSDK_USE_BUILDTIME_NAMESPACES_SET;
+
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
@@ -55,9 +58,22 @@ public class ConfigurationServiceRecorder {
                 c.setRetryConfiguration(null);
             }
 
-            // if despite all of this, we still haven't set the namespaces, use the operator-level default if it exists
+            // if the namespaces weren't set as an annotation, use the operator-level configuration if it exists
             if (!c.isWereNamespacesSet()) {
-                runTimeConfiguration.namespaces.ifPresent(ns -> c.setNamespaces(new HashSet<>(ns)));
+                // The namespaces field has a default value so that we are able to detect if the configuration value is set to "".
+                // Setting the value to "" will reset the configuration and result in an empty Optional.
+                // Not setting the value at all will result in the default being applied, which we can test for.
+                if (runTimeConfiguration.namespaces.isPresent()) {
+                    final var runtimeNamespaces = new HashSet<>(runTimeConfiguration.namespaces.get());
+                    // If it's not the default value, use it because it was set.
+                    // If it is the default value, ignore it and let any build time config be used.
+                    if (!QOSDK_USE_BUILDTIME_NAMESPACES_SET.equals(runtimeNamespaces)) {
+                        c.setNamespaces(runtimeNamespaces);
+                    }
+                } else {
+                    // Value has been explicitly reset (value was empty string), use all namespaces mode
+                    c.setNamespaces(DEFAULT_NAMESPACES_SET);
+                }
             }
         });
 
--- a/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java
+++ b/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java
@@ -14,6 +14,8 @@
  */
 package io.quarkiverse.operatorsdk.runtime;
 
+import static io.quarkiverse.operatorsdk.runtime.Constants.QOSDK_USE_BUILDTIME_NAMESPACES;
+
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


metacosm:
Looks reasonable, was thinking about doing something similar indeed. 

Would you mind changing the commit message so that it fits the conventional format, though, please?

csviri (approved):
LGTM

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java
```diff
@@ -55,9 +58,22 @@ public Supplier<QuarkusConfigurationService> configurationServiceSupplier(Versio
...
-            // if despite all of this, we still haven't set the namespaces, use the operator-level default if it exists
+            // if the namespaces weren't set as an annotation, use the operator-level configuration if it exists
             if (!c.isWereNamespacesSet()) {
-                runTimeConfiguration.namespaces.ifPresent(ns -> c.setNamespaces(new HashSet<>(ns)));
+                // The namespaces field has a default value so that we are able to detect if the configuration value is set to "".
+                // Setting the value to "" will reset the configuration and result in an empty Optional.
+                // Not setting the value at all will result in the default being applied, which we can test for.
+                if (runTimeConfiguration.namespaces.isPresent()) {
+                    final var runtimeNamespaces = new HashSet<>(runTimeConfiguration.namespaces.get());
+                    // If its not the default value, use it because it was set.
```
metacosm, line 68:
```suggestion
                    // If it's not the default value, use it because it was set.
```

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java
```diff
@@ -59,7 +61,10 @@ public class RunTimeOperatorConfiguration {
      * specified by the kube config file the operator uses.
      * </p>
      */
-    @ConfigItem
+    // We use a default here so that we are able to detect if the configuration value is set to "". Setting the value to "" will
+    // reset the configuration and result in an empty Optional, but not setting the value at all will result in the default being
+    // applied.
+    @ConfigItem(defaultValue = QOSDK_USE_BUILDTIME_NAMESPACES)
```
metacosm, line 67: I'm wondering if it might not just be enough to set the default value to `io.javaoperatorsdk.operator.api.reconciler.Constants.WATCH_ALL_NAMESPACES` here. Ideally, we'd need a test to cover this, actually.

metacosm (changes requested):
Not sure if we can't support this in a simpler way, a test case would help figure this out… Otherwise the general idea is the right one, I think.

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java
Jamstah: If we set the default to WATCH_ALL_NAMESPACES we couldn't tell if there was no specific runtime setting, which means we would _always_ override the build time property.

Maybe the build time property doesn't matter because it's only really intended to affect the manifests and not the runtime behaviour... but at the moment the code will honour the build time property if the runtime property is not set at all.

Jamstah:
I had a look at how to test this - at the moment the integration tests run a single operator with all the test controllers injected. If I were to set the global operator namespace property for those tests, it would break all the existing tests.

What are your thoughts on how to architect a test here? A second set of integration tests? A second profile? Something else?

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/RunTimeOperatorConfiguration.java
metacosm: > Maybe the build time property doesn't matter because it's only really intended to affect the manifests and not the runtime behaviour... but at the moment the code will honour the build time property if the runtime property is not set at all.

I would say that this is the intended behavior but I'm open to change my mind about this depending on what people think. The previous version used the same property for both build and runtimes, which was conducive to confusing which version was used ultimately. With the new `generate-with-watched-namespaces` property, it is now more explicit that the build time value is used mostly for generation purposes. Its value can indeed be propagated at runtime and it's indeed how things are currently implemented, just not sure if it actually make sense since it would make it more difficult to know exactly which configuration is used at runtime. Completely separating both might make more sense with that respect.

Jamstah: That was my thinking with a property that was explicitly namespaced with `manifests` in the name to separate it from properties that affect the built image.

I feel like the manifests are really runtime configuration and tbh, I think using quarkus.operator-sdk.namespaces at build time should affect the manifests and not the image, and that would make conceptual sense. It breaks the duplicated properties rule though.

metacosm:
> I had a look at how to test this - at the moment the integration tests run a single operator with all the test controllers injected. If I were to set the global operator namespace property for those tests, it would break all the existing tests.
> 
> What are your thoughts on how to architect a test here? A second set of integration tests? A second profile? Something else?

I'm thinking about specific tests using a [specific test profile](https://quarkus.io/guides/getting-started-testing#writing-a-profile).

metacosm:
Thank you. I will add tests in a separate PR.

Jamstah:
Thanks Chris :)

I did have a look at it, but think I'd need more time for the profiles thing!

PR #665, opened 2023-07-20
feat: generate simple generic helm chart

csviri:
This is the MVP for helm chart generator, it is "bare minimum" that works, will add subsequent PRs to improve.


core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/DisableDefaultHelmListener.java
```diff
@@ -0,0 +1,21 @@
+package io.quarkiverse.operatorsdk.deployment.helm;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.dekorate.WithSession;
+import io.dekorate.kubernetes.config.BaseConfigFluent;
+import io.dekorate.kubernetes.config.Configurator;
+
+public class DisableDefaultHelmListener extends Configurator<BaseConfigFluent<?>> implements WithSession {
```
metacosm, line 10: What's the point of this class?

samples/mysql-schema/src/main/resources/application.properties
```diff
@@ -4,4 +4,5 @@ quarkus.native.additional-build-args=--initialize-at-run-time=org.apache.commons
 quarkus.datasource.username=root
 quarkus.datasource.devservices.image-name=mariadb:10.7
 quarkus.kubernetes-client.devservices.override-kubeconfig=true
-quarkus.http.test-port=0
\ No newline at end of file
+quarkus.http.test-port=0
+quarkus.kubernetes.image-pull-policy=IfNotPresent
```
metacosm, line 8: Why was this added?

csviri: I think that was just to test with minikube locally, can remove for now

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/DisableDefaultHelmListener.java
csviri: added comment. it is to disable default helm generator

samples/mysql-schema/src/main/resources/application.properties
csviri: done

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/HelmConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class HelmConfiguration {
+
+    /**
+     * Can be used to disable helm chart generation.
+     */
+    @ConfigItem(defaultValue = "true")
```
metacosm, line 12: This should be false by default, imo. We don't want to generate Helm charts on every build for sure.

csviri: IDK, I would rather generate everything, users still can opt out, if something is not needed, but they would be aware that it's there. Also would worth to measure the time difference. Will do some naive measurement

metacosm: This is more important in the dev mode, so at the very least, helm chart generation should be disabled in dev mode, though I still think it shouldn't be active by default, the same way we don't generate bundles by default either (which is why the support is in a different module). Since helm support is in the core module, it should be deactivated by default. Generating something all the time just to make people aware of it is a bad idea, imo. This should be addressed by documentation.

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/HelmChartProcessor.java
```diff
@@ -0,0 +1,224 @@
...
+            "service.yaml",
+            "serviceaccount.yaml"
+    };
+    public static final String CHART_YAML_FILENAME = "Chart.yaml";
+    public static final String VALUES_YAML_FILENAME = "values.yaml";
+    public static final String CRD_DIR = "crds";
+
+    @BuildStep
+    public void handleHelmCharts(
+            BuildProducer<ArtifactResultBuildItem> dummy,
```
metacosm, line 59: If this is needed to ensure this steps runs after another one, please add a comment to that effect, otherwise that parameter might get removed at some point, leading to possibly difficult to diagnose issues.

metacosm: Actually, this is a BuildProducer that isn't used so I don't think there's any point in keeping this.

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/HelmConfiguration.java
csviri: IDK, can do it for now, but my opinion is that if there is no explicit reason why not to generate it it should be there, even for dev mode, so users can see the helm chart if they add an additional dependent resource. 

Why bundle generation is a separate module, I mean why it is not just a flag to not generate it?

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/HelmChartProcessor.java
csviri: without it it does not gets executed

csviri: maybe there is a more elegant way to do this

core/deployment/src/test/java/io/quarkiverse/operatorsdk/test/HelmChartGeneratorTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class HelmChartGeneratorTest {
+
+    @RegisterExtension
+    static final QuarkusProdModeTest config = new QuarkusProdModeTest()
+            .setApplicationName("helm-chart-test")
+            .withApplicationRoot(
+                    (jar) -> jar.addClasses(SimpleReconciler.class, SimpleCR.class, SimpleSpec.class, SimpleStatus.class));
+
+    @Test
+    void generatesHelmChart() {
```
Sgitario, line 21: I think it would be good to assert that the resources are indeed generated.

Sgitario (approved):
I see that you have followed the approach of using templates instead of YAML expressions which is perfectly fine and I think it fits better in your use case. 

I've added a few comments about the tests, but not blockers at all. 

Good job!!

core/deployment/src/test/java/io/quarkiverse/operatorsdk/test/HelmChartGeneratorTest.java
csviri: yes, will take a look on this, there was an issue with the test, and though an e2e will be enough, bit will give it an other try. Thx!

samples/exposedapp/src/test/java/io/halkyon/HelmDeploymentE2ETest.java
```diff
@@ -0,0 +1,194 @@
...
+import io.javaoperatorsdk.operator.Operator;
+import io.javaoperatorsdk.operator.api.reconciler.Constants;
+import io.quarkus.test.junit.QuarkusTest;
+
+// currently there is a bug with kind testcontainer that blocks the e2e tests done properly
+// will finish this test when it gets resolved:
+// https://github.com/dajudge/kindcontainer/issues/235
+@Disabled
+@QuarkusTest
+class HelmDeploymentE2ETest {
```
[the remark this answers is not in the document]
csviri: Yeah, it is quite annoying when you need those tools, but managing such tests is quite handy from java code. Will take a look on this in next pr. An other possibility is to commit something like this script to the repo:
https://helm.sh/docs/intro/install/#from-script
And install helm during the test if not installed yet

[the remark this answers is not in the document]
csviri: For GitHub Actions there is a nice action that does that.

core/deployment/src/test/java/io/quarkiverse/operatorsdk/test/HelmChartGeneratorTest.java
csviri: added simple asserts, not sure if we should check the templates in depth in this case.

core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/HelmConfiguration.java
metacosm: > IDK, can do it for now, but my opinion is that if there is no explicit reason why not to generate it it should be there, even for dev mode, so users can see the helm chart if they add an additional dependent resource.

The core principle is to be as efficient as possible so sure, we can always say things are "fast" enough but from my point of view, if something doesn't need to be done all the time, it shouldn't.

> 
> Why bundle generation is a separate module, I mean why it is not just a flag to not generate it?

Cleaner separation of concerns. Also allows to not bring the dependencies if not used. For that matter, we might want to extract Helm support in its own module at some later point as well.

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/HelmChartProcessor.java
```diff
@@ -0,0 +1,219 @@
...
+        });
+
+    }
+
+    @SuppressWarnings("rawtypes")
+    private void addPrimaryClusterRoleBindings(File helmDir, Collection<QuarkusControllerConfiguration> controllerConfigs) {
+        try (InputStream file = Thread.currentThread().getContextClassLoader()
+                .getResourceAsStream(CRD_ROLE_BINDING_TEMPLATE_PATH)) {
+            if (file == null) {
+                throw new IllegalArgumentException("Template file " + CRD_ROLE_BINDING_TEMPLATE_PATH + " doesn't exist");
```
csviri, line 120: There is not much benefit to check for null on a static resource.

metacosm (approved):

PR #669, opened 2023-07-27
feat: add config option for informerListLimit

csviri:


csviri:
@metacosm  do I understand correctly that the error: 
```
Caused by: java.lang.ClassCastException: class java.util.Optional cannot be cast to class java.lang.Long (java.util.Optional and java.lang.Long are in module java.base of loader 'bootstrap')
```

is because of the Optional, is there an other way to fix this, than rename the property in this case?

metacosm:
LGTM but will keep this open until https://github.com/operator-framework/java-operator-sdk/pull/2004 is available in a JOSDK release so that we add the associated changes here as well.

csviri:
@metacosm kk, will prepare those changes

metacosm (approved):

PR #686, opened 2023-08-16
feat: provide a CSV.spec.version value when possible

### LINKED ISSUE #682: ClusterServiceVersion.spec.version might be required
Just for reference. Maybe this will be an issue, soon.
https://github.com/operator-framework/operator-registry/issues/1135

metacosm:
Fixes #682

### DIFF
--- a/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/Utils.java
+++ b/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/Utils.java
@@ -4,9 +4,12 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
+import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.Assertions;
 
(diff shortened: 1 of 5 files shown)


bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/Utils.java
```diff
@@ -4,9 +4,12 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 
+import org.jetbrains.annotations.NotNull;
```
csviri, line 7: This import probably should not be here

csviri (approved):
Just found one strageness with imports, otherwise LGTM, although will need to dig deeper in this part

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/Utils.java
metacosm: Good catch!

PR #703, opened 2023-09-04
feat: make it possible to mark a primary resource as unowned

metacosm:


core/deployment/src/test/java/io/quarkiverse/operatorsdk/test/UnownedTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class UnownedTest {
+    // Start unit test with your extension loaded
+    @RegisterExtension
+    static final QuarkusProdModeTest config = new QuarkusProdModeTest()
+            .setApplicationName(UnownedReconciler.NAME)
+            .withApplicationRoot(
+                    (jar) -> jar.addClasses(UnownedReconciler.class, TestCR.class))
+            .overrideConfigKey("quarkus.operator-sdk.crd.generate-all", "true")
+            .overrideConfigKey("quarkus.operator-sdk.controllers." + UnownedReconciler.NAME + ".unowned-primary", "true");
```
metacosm, line 24: This is the salient part of this test: setting the `unowned-primary` property to `true` for the controller in application.properties.

PR #724, opened 2023-09-22
feat: use package name in olm bundle's metadata

### LINKED ISSUE #723: Bundle generator overwrites package name with bundle name
The budnle generator sets the value of `operators.operatorframework.io.bundle.package.v1` in `metadata/annotation.yaml` to the name of the bundle. 

There is a 1-to-M  relation between OLM package and OLM bundle. This means that the package is most likely going to be static between new bundle releases of the same operator. While the Bundle itself is valid it is impossible to generate multiple such bundles into the same catalog without overwriting the default into in bundles metadata (which is confusing and will cause issue for example when contributing such bundle to operatorhub)

jcechace:
fixes #723


metacosm (approved):
Thank you!

PR #726, opened 2023-09-23
fix: runtime namespaces are now correctly set from controller properties 

### LINKED ISSUE #725: Runtime namespaces are not set if annotation sets namespaces during build time
The condition [here](https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java#L61-L77) is ultimately always false when `ControllerConfig` annotation sets namespaces (see [here](https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/QuarkusControllerConfigurationBuildStep.java#L219C29-L229))

Additionally the current implementation takes into account only the operator level property so  any `QUARKUS_OPERATOR_SDK_CONTROLLERS_<NAME>_NAMESPACES` won't have any effect it seems.

Note: Removing `ControllerConfig.namespaces` from the reconcilier will make the operator level runtime config work as expected. However this leads to another set of issues which I'm gonna skip to keep this report cleaner.

jcechace:
Fixes #725

@metacosm  If there are soem tests which I can draw inspiration from, let me know and I will add some coverage. Otherwise I will be leaving this up to you as understanding how exactly I should write the tests is something I don't have the time for right now.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java
+++ b/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/ConfigurationServiceRecorder.java
@@ -4,6 +4,7 @@ import static io.javaoperatorsdk.operator.api.reconciler.Constants.DEFAULT_NAMES
 import static io.quarkiverse.operatorsdk.runtime.Constants.QOSDK_USE_BUILDTIME_NAMESPACES_SET;
 
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -45,8 +46,8 @@ public class ConfigurationServiceRecorder {
             if (extConfig != null) {
                 extConfig.finalizer.ifPresent(c::setFinalizer);
                 extConfig.selector.ifPresent(c::setLabelSelector);
-                extConfig.namespaces.map(HashSet::new).ifPresent(c::setNamespaces);
                 c.setRetryConfiguration(RetryConfigurationResolver.resolve(extConfig.retry));
+                setNamespacesFromRuntime(c, extConfig.namespaces);
             }
 
             // set retry to default if it hasn't been set already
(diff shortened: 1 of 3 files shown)


jcechace:
@metacosm This PR respects the precendence order of  `controller level prop` -> `annotation` -> `opertor level (blanket) prop`. However I still think that such priority of evaluation is illogical and inherently flawed for runtime use. There are several reasons why


1) Annotation is a build time configuration by nature, it doesn't make sense that a runtime property (albeit a global / blanekt one) has lower precendence 
2) Annotation value  (being a build time config) is hidden from the actual person deploying/using the operator. It taking precendence over something explicitly visible (such as `QUARKUS_OPERATOR_SDK_NAMESPACES`) is IMHO extremely confusing. 


Personally I would see the order as`controller level prop` -> `opertor level (blanket) prop` -> `annotation`. 

The reasoning in this case is that `QUARKUS_OPERATOR_SDK_NAMESPACES` is equivalent to setting `QUARKUS_OPERATOR_SDK_CONTROLLERS_<name>_NAMESPACES` for each controller. The ability to override the setting for specific controller on runtime is of course kept.

metacosm:
> @metacosm This PR respects the precendence order of `controller level prop` -> `annotation` -> `opertor level (blanket) prop`. However I still think that such priority of evaluation is illogical and inherently flawed for runtime use. There are several reasons why
> 
>     1. Annotation is a build time configuration by nature, it doesn't make sense that a runtime property (albeit a global / blanekt one) has lower precendence
> 
>     2. Annotation value  (being a build time config) is hidden from the actual person deploying/using the operator. It taking precendence over something explicitly visible (such as `QUARKUS_OPERATOR_SDK_NAMESPACES`) is IMHO extremely confusing.
> 
> 
> Personally I would see the order as`controller level prop` -> `opertor level (blanket) prop` -> `annotation`.
> 
> The reasoning in this case is that `QUARKUS_OPERATOR_SDK_NAMESPACES` is equivalent to setting `QUARKUS_OPERATOR_SDK_CONTROLLERS_<name>_NAMESPACES` for each controller. The ability to override the setting for specific controller on runtime is of course kept.

Yes, this makes sense, however I will not make this breaking change (though you could argue that this is currently broken anyway) in a patch release. We will change this behavior to follow the one you describe in the next minor release.

jcechace:
> > @metacosm This PR respects the precendence order of `controller level prop` -> `annotation` -> `opertor level (blanket) prop`. However I still think that such priority of evaluation is illogical and inherently flawed for runtime use. There are several reasons why
> > ```
> > 1. Annotation is a build time configuration by nature, it doesn't make sense that a runtime property (albeit a global / blanekt one) has lower precendence
> > 
> > 2. Annotation value  (being a build time config) is hidden from the actual person deploying/using the operator. It taking precendence over something explicitly visible (such as `QUARKUS_OPERATOR_SDK_NAMESPACES`) is IMHO extremely confusing.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Personally I would see the order as`controller level prop` -> `opertor level (blanket) prop` -> `annotation`.
> > The reasoning in this case is that `QUARKUS_OPERATOR_SDK_NAMESPACES` is equivalent to setting `QUARKUS_OPERATOR_SDK_CONTROLLERS_<name>_NAMESPACES` for each controller. The ability to override the setting for specific controller on runtime is of course kept.
> 
> Yes, this makes sense, however I will not make this breaking change (though you could argue that this is currently broken anyway) in a patch release. We will change this behavior to follow the one you describe in the next minor release.

Sure thing -- that was the primary reason for not including it here. I can follow up with another PR later.

metacosm:
Merging this, will change the behavior in the next minor and add tests there.

jcechace:
@metacosm I think this should have also gotten into 6.4 and 6.5 (so main maybe?)

metacosm:
> @metacosm I think this should have also gotten into 6.4 and 6.5 (so main maybe?)

Indeed, I never got around to making the changes I initially intended so will add this change to `main` as well.

PR #734, opened 2023-10-03
feat: add Quarkus CLI commands

metacosm:
- chore: initial work on CLI plugin
- wip: first commands layout
- feat: first implementation of API creation command

### DIFF
--- a/common/src/main/java/io/quarkiverse/operatorsdk/common/Files.java
+++ b/common/src/main/java/io/quarkiverse/operatorsdk/common/Files.java
@@ -1,0 +1,165 @@
+package io.quarkiverse.operatorsdk.common;
+
+import java.io.*;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.*;
+import java.nio.file.spi.FileSystemProvider;
+import java.util.Collections;
+import java.util.Map;
+
+import io.quarkus.qute.Qute;
+import io.quarkus.qute.TemplateException;
+import io.quarkus.qute.TemplateInstance;
+
+public class Files {
+
+    private static final String API_TEMPLATES_PATH = "templates/cli/api";
+    private static final String ROOT_TARGET_DIR = "src/main/java/";
+
+    private Files() {
+    }
+
+    public interface MessageWriter {
+        default void write(String message) {
+            write(message, null, false);
+        };
+
+        default void error(String message) {
+            error(message, null);
+        }
+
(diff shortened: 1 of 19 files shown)


csviri (approved):
LGTM

common/src/main/java/io/quarkiverse/operatorsdk/common/Files.java
```diff
@@ -0,0 +1,165 @@
...
+                "version", version,
+                "kind", finalKind);
+
+        final var apiTemplatesPath = Paths.get(apiTemplatesDirURI);
+        try (var paths = java.nio.file.Files.walk(apiTemplatesPath)) {
+            paths.filter(f -> java.nio.file.Files.isRegularFile(f) && f.getFileName().toString().endsWith(".qute"))
+                    .forEach(path -> {
+                        try {
+                            final var templateAsString = java.nio.file.Files.readString(path);
+                            final var templateInstance = Qute.fmt(templateAsString)
```
mkouba, line 103: Like I mentioned, this is not ideal in the sense that the template content is parsed every time the `Qute#fmt()` method is called. If you need to render the template multiple times then either use `Qute.enableCache()` or Qute.engine().parse(...)` together with `Qute.engine().putTemplate()` and `Qute.engine().getTemplate()`...

mkouba: What I'm saying is: if you only render the template once, then `Qute.fmt(templateAsString)` is OK. If multiple times, then either use `Qute.enableCache()` - this enables cache for `Qute.fmt()` by default (`templateAsString` will be parsed once); or use something like:
```java
// This should be called once
Engine engine = Qute.engine();
engine.putTemplate("myTemplate", engine.parse(templateAsString));
...
// And then to render a template:
Qute.engine().getTemplate("myTemplate").render();
```

metacosm: Ah, I think I was missing the `putTemplate` part, I thought that I could call `getTemplate` directly and that wasn't working. Thanks! That said, since these things won't probably be run very often, I'm not sure it matters but I will still make the change (it might help when called from the dev console, maybe?). Then again, you could make a point that if it's not going to be call often, caching a template might actually be worse because then you commit memory to something that will only be used infrequently and possibly only once.

mkouba: Yes, it really depends on the use case :shrug: .

PR #782, opened 2023-12-12
feat: parallelize Helm chart generation

metacosm:


metacosm:
> I see you try to remove dekorate, note that with this it won't be possible nicely customize the Deployment as it is now.

Not sure I'm following. Would you mind elaborating, please?

csviri:
> > I see you try to remove dekorate, note that with this it won't be possible nicely customize the Deployment as it is now.
> 
> Not sure I'm following. Would you mind elaborating, please?

Basically here I'm using the deployment:
https://github.com/quarkiverse/quarkus-operator-sdk/blob/4fd5772acbe18ee6ff64e91803e77f05629a622a/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/HelmChartProcessor.java#L96-L96

From generated `kubernetes.yaml` from dekorate, and that is highly customizable using application.properties. if we are going to generate that from scratch, the whole customization mechanism going away that is inherited from dekorate.

csviri (approved):
LGTM

Issue #801, opened 2024-01-17
Helm chart generation incorrectly filters resources from kubernetes.yaml


jcechace:
`HelmChartProcessor` seems to rely on `ApplicationInfoBuildItem` for quite a few value. For example here 

https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/helm/HelmChartProcessor.java#L158

However I don't think that  `ApplicationInfoBuildItem#name` is the expected value here. In my opinion the correct value is  the one set to  `quarkus.kubernetes.name` (and `quarkus.kubernetes.version` for version respectively). The `quarkus.kubernetes.name` property actually defaults to application name. 

The obvious workaround is to also set `quarkus.application.name`.  However when setting both 

`quarkus.application.name=debezium-operator` and
`quarkus.kubernetes.name=debezium-operator` 

together with the following `src/kubernetes/kubernetes.yml`

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: config-view
rules:
  - apiGroups:
      - ""
    resources:
      - secrets
      - configmaps
    verbs:
      - get
      - list
      - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: debezium-operator-config-view
roleRef:
  kind: ClusterRole
  apiGroup: rbac.authorization.k8s.io
  name: config-view
subjects:
  - kind: ServiceAccount
    name: debezium-operator
```

Resulted in no `kubernetes.yml` being generated in the chart's templates directory. Maybe I'm missing something but the filtering logic should let these through

jcechace:
@csviri @metacosm  when we are at it. Looking at our contents of `kubernetes.yaml` is there perhaps a better way to add additional rules to the cluster role which is already generated for the operator sa?

metacosm:
> @csviri @metacosm when we are at it. Looking at our contents of `kubernetes.yaml` is there perhaps a better way to add additional RBAC rules to the cluster role which is already generated for the operator sa?

You can add additional `PermissionRules` to the generated RBACs as seen at https://github.com/quarkiverse/quarkus-operator-sdk/blob/e646ede9b4fca166c262a4afd66eb3003f3137fb/samples/joke/src/main/java/io/quarkiverse/operatorsdk/samples/joke/JokeRequestReconciler.java#L36.
[...]

PR #805, opened 2024-01-23
fix: use Kubernetes configuration for name and version when available

metacosm:
- fix: use Kubernetes configuration for name and version when available
- refactor: RBACRule should be used instead of PermissionRule

### DIFF
--- a/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
+++ b/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
@@ -105,8 +105,11 @@ public @interface CSVMetadata {
         String[] verbs() default { "get", "list", "watch", "create", "delete", "patch", "update" };
 
         /**
-         * @return the service account name for the permission rule. If not provided, it will use the default service account
-         *         name.
+         * @return the service account name to which the permission rule will be assigned. If not provided, the default service
+         *         account name as defined for your operator will be used. Note that for the rule to be effectively added to the
+         *         CSV, a service account with that name <em>must</em> exist in the generated kubernetes manifests as this is
+         *         the base upon which the bundle generator works. This means that if you add a rule that targets a service
+         *         account that is not present in the generated manifest, then this rule won't appear in the generated CSV.
          */
         String serviceAccountName() default "";
     }
--- a/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
+++ b/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
@@ -91,6 +91,12 @@ public @interface CSVMetadata {
         boolean supported() default true;
     }
 
+    /**
+     * Additional RBAC rules that need to be provided because they cannot be inferred automatically. Note that RBAC rules added
+     * to your reconciler via {@link RBACRule} should already be handled automatically, under the service account name
+     * associated with your Reconciler so this annotation should only be used to add additional rules to other service accounts
+     * or for rules that you don't want to appear in the generated Kubernetes manifests.
+     */
     @interface PermissionRule {
         String[] apiGroups();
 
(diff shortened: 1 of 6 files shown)


annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
```diff
@@ -91,6 +91,12 @@
         boolean supported() default true;
     }
 
+    /**
+     * Additional RBAC rules that need to be provided because they cannot be inferred automatically. Note that RBAC rules added
+     * to your reconciler via {@link RBACRule} should already be handled automatically, under the service account name
+     * associated with your Reconciler so this annotation should only be used to add additional rules to other service accounts
+     * or for rules that you don't want to appear in the generated Kubernetes manifests.
+     */
```
jcechace, line 99: Kudos for this. Javadocs should be used way more :) 

@metacosm one thing I am still curious about is this being part of the `CSVMetadata`. Shouldn't this also have an impact when CSV bundle is not used? AFAIK the `CSVMetadata` class really is used just for CSV metadata generation, or am I wrong?

metacosm: `RBACRule` is not part of `CSVMetadata` and we still need permission rules on `CSVMetadata` in case you don't want something to be generated in the kubernetes manifests. Or am I misunderstanding your question?

jcechace:
@metacosm  

```java
/**
         * @return the service account name for the permission rule. If not provided, it will use the default service account
         *         name.
         */
        String serviceAccountName() default "";    
```
Although this is not affected by this PR I have a question/note to add here. What is considered a "default service account"? When I incorrectly used `CSVMetadata.permissionRules` with the `@PermissionRule` (instead of `@RBACRule`directly on the Reconciler) and haven't provided this `serviceAccountName` parameter it didn't seem to generate anything anywhere.

metacosm:
> @metacosm
> 
> ```java
> /**
>          * @return the service account name for the permission rule. If not provided, it will use the default service account
>          *         name.
>          */
>         String serviceAccountName() default "";    
> ```
> 
> Although this is not affected by this PR I have a question/note to add here. What is considered a "default service account"? When I incorrectly used `CSVMetadata.permissionRules` with the `@PermissionRule` (instead of `@RBACRule`directly on the Reconciler) and haven't provided this `serviceAccountName` parameter it didn't seem to generate anything anywhere.

Yes, this is confusing and something that I'm pondering. I will add some documentation here.

jcechace:
> > @metacosm
> > ```java
> > /**
> >          * @return the service account name for the permission rule. If not provided, it will use the default service account
> >          *         name.
> >          */
> >         String serviceAccountName() default "";    
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Although this is not affected by this PR I have a question/note to add here. What is considered a "default service account"? When I incorrectly used `CSVMetadata.permissionRules` with the `@PermissionRule` (instead of `@RBACRule`directly on the Reconciler) and haven't provided this `serviceAccountName` parameter it didn't seem to generate anything anywhere.
> 
> Yes, this is confusing and something that I'm pondering. I will add some documentation here.

Regarding your comment (https://github.com/quarkiverse/quarkus-operator-sdk/issues/801#issuecomment-1906421052) 

I'm confused how these two relate together. We are still using QOSDK 6.3.5 which obviously doesn't  include this fix. However the permission is added to the CSV just fine when only `@RBACRule` is used

annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java
jcechace: > `RBACRule` is not part of `CSVMetadata` and we still need permission rules on `CSVMetadata` in case you don't want something to be generated in the kubernetes manifests. Or am I misunderstanding your question?
I see.. That makes sense. So you would use this when you don't want the permission in your `kubernetes.yml` but still want to use it for CSV. Thanks!

PR #812, opened 2024-01-26
fix: helm chart generates invalid YAML when watching all namespaces

adutra:
When a reconciler is configured to watch all namespaces, a ClusterRoleBinding object is generated by
AddRoleBindingsDecorator, and it later gets appended to templates/kubernetes.yml by HelmChartProcessor.

However, that file becomes syntactically invalid because FileUtils.asYaml produces a top-level YAML array when passed a List. That's not what Helm expects.

When one tries to install the generated chart, the following error is returned:

     Error: INSTALLATION FAILED: YAML parse error on
     [...]/templates/kubernetes.yml: error unmarshaling JSON:
     while decoding JSON: json: cannot unmarshal array into
     Go value of type releaseutil.SimpleHead
     YAML parse error on [...]/templates/kubernetes.yml

### DIFF
--- a/common-deployment/src/main/java/io/quarkiverse/operatorsdk/common/FileUtils.java
+++ b/common-deployment/src/main/java/io/quarkiverse/operatorsdk/common/FileUtils.java
@@ -2,9 +2,7 @@ package io.quarkiverse.operatorsdk.common;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.util.List;
 
-import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.client.utils.KubernetesSerialization;
 
 public class FileUtils {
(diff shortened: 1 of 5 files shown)


adutra:
@csviri I wonder if the ClusterRoleBinding doc shouldn't be filtered out by `HelmChartProcessor#filterOutStandardResources`? It seems we already have an equivalent ClusterRoleBinding being generated in the `xyz-reconciler-crd-role-binding.yaml` file.

adutra:
I thought some more about this and figured that it made sense to exclude the generated cluster role binding.

I added another commit that excludes it, and enhances the test to produce some custom resources and check that they all get appended to kubernetes.yml.

csviri (approved):
Makes sense, and LGMT

metacosm (approved):
Thank you for your contribution!

PR #822, opened 2024-02-07
improve: use chart appVersion in labels

csviri:
drops redundant `Values.version`


metacosm (approved):
Looks OK to me but I'll let @jcechace give the last word on it.

csviri:
Will do also some additional testing before merge

PR #840, opened 2024-03-04
feat(ci): use latest OLM version

metacosm:


xstefank (approved):

metacosm:
Reasons of why OLM v1 doesn't work should be investigated.

xstefank:
I don't remember what was the issue with this but it seems to work now. Do you want to limit the version again or test both OLM latest v1 and 0.x?

metacosm:
Merging, let's see how things go moving forward.

PR #841, opened 2024-03-04
fix: add proper default install mode derived from watched namespaces

### LINKED ISSUE #839: CSV InstallMode is not correct by default and should take namespaces information into account
Right now, the install mode is only changed if CSVMetadata is provided and defaults to `AllNamespaces`, which isn't correct.

metacosm:
Fixes #839

### DIFF
--- a/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/MultipleOperatorsBundleTest.java
+++ b/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/MultipleOperatorsBundleTest.java
@@ -1,8 +1,7 @@
 package io.quarkiverse.operatorsdk.bundle;
 
 import static io.quarkiverse.operatorsdk.bundle.Utils.*;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.IOException;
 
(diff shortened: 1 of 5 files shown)


csviri (approved):

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/FirstReconciler.java
```diff
@@ -5,7 +5,7 @@
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
 
-@CSVMetadata(name = "first-operator", version = FirstReconciler.VERSION)
+@CSVMetadata(name = "first-operator", version = FirstReconciler.VERSION, installModes = @CSVMetadata.InstallMode(type = "MultiNamespace", supported = false))
```
xstefank, line 8: Maybe there should be a public wrapper around the allowed type values.

xstefank (approved):

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/MultipleOperatorsBundleTest.java
```diff
@@ -49,6 +48,13 @@ public void shouldWriteBundleForTheOperators() throws IOException {
         assertEquals(FirstReconciler.REPLACES, csv.getSpec().getReplaces());
         var bundleMeta = getAnnotationsFor(bundle, "first-operator");
         assertEquals(BUNDLE_PACKAGE, bundleMeta.getAnnotations().get("operators.operatorframework.io.bundle.package.v1"));
+        assertEquals(2, csv.getSpec().getInstallModes().size());
+        var installMode = csv.getSpec().getInstallModes().get(0);
+        assertEquals("AllNamespaces", installMode.getType());
```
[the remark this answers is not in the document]
metacosm: I'm always debating whether or not to use constants in the tests, especially when the tests are checking compatibility with an external service or spec. In this case, if we used a constant, and the constant value changed **but not the external spec**, then the test would still pass even though it would now output incorrect values.

bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/sources/FirstReconciler.java
metacosm: Do you mean like an enum or something like that?

metacosm:
Looks like the install mode is not supported: https://github.com/quarkiverse/quarkus-operator-sdk/actions/runs/10471449036/job/28998782965#step:32:134

xstefank:
@metacosm do you want to close this PR then?

metacosm:
Obsolete, to be revisited if needed.

Issue #738, opened 2023-10-16
CSVMetadata.name is confusing in some cases (e.g. when providing meta information for CRDs


jcechace:
`CSVMetadata.name` is used for several things currently

1. To set the bundle name
2. To group together various objects related to the same bundle

These points are in some way problematic

- A bundle name usually contains a version which changes over time, however the "grouping" identifier would ideally be static 
- In some cases the `CSVMetadata` annotation as a whole is used to provide metadata for just a part of the resulting `CSV`. Such is the case for `.spec.customresourcedefinitions.owned[]` property. However the CRD definitions in this case also contain a `name` attribute. While it doesn't make sense to change the value of this attribute, it is confusing when one uses the `CSVMetadata.name` over the CR Java class in order to group the metadata into the correct bundle. 

Related [discussion on Discord](https://discord.com/channels/723455000604573736/786172306804047932/1159503558836748328)

**Proposal:**

Introduce `CSVMetadata.bundle` which would act as the "grouping" identifier and have ``CSVMetadata.name` default to `CSVMetadata.bundle` when processing the top level `CSVMetadata` annotation (that is the instance placed over either the controller class or class implementing `SharedCSVMetadata`).

It might also be worth considering whether metadata for specific parts of `CSV` (e.g. the `customresourcedefinitions` should be set using the generic `CSVMetadata` annotation as in this case the majority of its attributes does not apply -- however I don't consider this to be a major issue if the above confusion is removed.

metacosm:
Let me know what you think about the PR, see if this improves things.

PR #851, opened 2024-03-15
feat: improve OLM support

metacosm:
- **feat: automatically add provider name and URL based on user and git repo**
- **feat(ci): validate all bundles except Joke that OLM v1 doesn't like**
- **feat: split name into bundle and csv names to avoid confusion**
- **fix: sanitize version**

### DIFF
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
@@ -140,6 +140,7 @@ public class BundleProcessor {
             if (scmInfo != null) {
                 var origin = scmInfo.getRemote().get("origin");
                 if (origin != null) {
+                    System.out.println(origin);
                     try {
                         int atSign = origin.indexOf('@');
                         if (atSign > 0) {
@@ -149,7 +150,7 @@ public class BundleProcessor {
                         }
 
                         int dotGit = origin.indexOf(".git");
-                        if (dotGit < origin.length() - 1) {
+                        if (dotGit > 0 && dotGit < origin.length() - 1) {
                             origin = origin.substring(0, dotGit);
                         }
                         return origin;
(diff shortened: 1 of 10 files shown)


csviri (approved):

metacosm:
Fixes #738

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -127,6 +139,37 @@ CSVMetadataBuildItem gatherCSVMetadata(KubernetesConfig kubernetesConfig,
         return new CSVMetadataBuildItem(csvGroups);
     }
 
+    private static String getDefaultProviderURLFromSCMInfo(ApplicationInfoBuildItem appConfiguration,
+            JarBuildItem jarBuildItem) {
+        final var maybeProject = KubernetesCommonHelper.createProject(appConfiguration, Optional.empty(),
```
jcechace, line 144: @metacosm need to be extremely careful here. This class uses Dekorate to get the URL info directly from local git config. Personally I consider this unexpected (compared to using VCS info from pom configuration) and it was also the source of the recent Quarkus CVE .

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -127,6 +139,37 @@ CSVMetadataBuildItem gatherCSVMetadata(KubernetesConfig kubernetesConfig,
...
+                jarBuildItem.getPath());
+        return maybeProject.map(project -> {
+            final var scmInfo = project.getScmInfo();
+            if (scmInfo != null) {
+                var origin = scmInfo.getRemote().get("origin");
```
jcechace, line 149: Again.. same as with that CVE. We should make sure that this can be explicitly overridable -- there are cases when origin might point actually be a fork which one doesn't want to publicly advertise  (e.g. when builds are for whatever reason using a mirror URL or when you simply have manual steps in your workflow)

Naros: I absolutely must echo @jcechace concern here.  In almost all use cases, `origin` for me is my fork, and `upstream` refers to the public repository.

metacosm: Yes, it's very tricky so maybe I'll just remove this and let people deal with it manually instead…

metacosm: > Again.. same as with that CVE. We should make sure that this can be explicitly overridable -- there are cases when origin might point actually be a fork which one doesn't want to publicly advertise (e.g. when builds are for whatever reason using a mirror URL or when you simply have manual steps in your workflow)

Note that this is for the default value and is overridable using `@CSVMetadata` and its `provider` field. That said, this computed value might be faulty in many situations so it's best to remove it.

PR #856, opened 2024-03-20
fix: remove duplicated and wrong env variable

metacosm:
- **fix: properly log when helm generation is disabled**
- **fix: do not modify deployment as it is live and reused elsewhere**


csviri (approved):
LGTM

jcechace:
@metacosm LGTM. I will attempt to build QOSDK from sources and  try it against real source. 

If everything works, is there a chance for you to cut a release containing this fix later today / tomorrow?

metacosm:
> @metacosm LGTM. I will attempt to build QOSDK from sources and try it against real source.
> 
> If everything works, is there a chance for you to cut a release containing this fix later today / tomorrow?

That's the plan.

jcechace:
@metacosm Just tested the fix and all seems to be in order. Thanks for the quick fix!

PR #894, opened 2024-07-02
feat: configure max reconciliation interval via property

### LINKED ISSUE #893: Configuring `maxReconciliationInterval` at runtime
I couldn't find a way to set/override `maxReconciliationInterval` via a config property. It would be useful to allow an installer to override this at runtime when the value given in `@ControllerConfiguration` isn't suitable for a particular environment.

metacosm:
Fixes #893


MikeEdgar (approved):
Looks good, thanks @metacosm

csviri (approved):

PR #920, opened 2024-08-12
refactor: switch BundleGenerationConfiguration to ConfigMapping

metacosm:

### DIFF
--- a/bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/BundleGenerationConfiguration.java
+++ b/bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/BundleGenerationConfiguration.java
@@ -4,28 +4,29 @@ import java.util.List;
 import java.util.Optional;
 
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
-import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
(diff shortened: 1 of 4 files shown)


bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/BundleGenerationConfiguration.java
```diff
@@ -4,49 +4,47 @@
...
 
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
-import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
 
-@ConfigRoot(name = "operator-sdk.bundle")
-public class BundleGenerationConfiguration {
+@ConfigMapping(prefix = "operator-sdk.bundle")
```
gsmet, line 11:
```suggestion
@ConfigMapping(prefix = "quarkus.operator-sdk.bundle")
```

gsmet (changes requested):
Spotted an issue. Not sure if it's going to fix your problem.

Also if you're not always generating a BuildItem that will match, you need to inject it with `Optional<YourBuildItem>`.

bundle-generator/runtime/src/main/java/io/quarkiverse/operatorsdk/bundle/runtime/BundleGenerationConfiguration.java
metacosm: Thanks!

PR #924, opened 2024-08-14
feat: use run_name to output workflow parameters

xstefank:


metacosm (approved):
Nice, wasn't aware of this feature!

PR #927, opened 2024-08-20
feat: update to v2 version of Fabric8's CRD generator

metacosm:


jcechace:
@metacosm  @xstefank  unfortunately in case default value is  specified for a property v2 currently may render invalid CRD as the value for "default" field in the schema is always going to be String regardless of the type value.

metacosm:
I will see if we can configure which version of the CRD generator to use. Can you use the 6.9.0 version that just got released, targeting Quarkus 3.17 or would this need to be backported to 6.8?

PR #931, opened 2024-08-22
Support typeless kubernetes resources when generating manifests

chendouble:
Typeless kubernetes resource has no Group and Plural value, so generating wrong policy rule.

Commit summary:
1) Use ApiGroup value from GVK and '*' for Plural value if dependent resource is assignable from GenericKubernetesDependentResource 2) Merge verbs in rule for optimization if with same ApiGroups and Resources

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddClusterRolesDecorator.java
+++ b/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddClusterRolesDecorator.java
@@ -133,6 +133,12 @@ public class AddClusterRolesDecorator extends ResourceProvidingDecorator<Kuberne
         return clusterRoleBuilder.build();
     }
 
+    /**
+     * Remove duplicated rules with same groups and resources, from which merge all verbs
+     *
+     * @param collectedRules may contain duplicated rules with same groups and resources, but different verbs
+     * @return no duplicated rules
+     */
     @NotNull
     private static Set<PolicyRule> getNormalizedRules(Set<PolicyRule> collectedRules) {
         Set<PolicyRule> normalizedRules = new LinkedHashSet<>();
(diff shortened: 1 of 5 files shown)


core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddClusterRolesDecorator.java
```diff
@@ -100,17 +121,59 @@ public static ClusterRole createClusterRole(QuarkusControllerConfiguration<?> cr
...
-        clusterRoleBuilder.addAllToRules(cri.getAdditionalRBACRules());
+        collectedRules.addAll(cri.getAdditionalRBACRules());
+        Set<PolicyRule> normalizedRules = getNormalizedRules(collectedRules);
+        clusterRoleBuilder.addToRules(normalizedRules.toArray(PolicyRule[]::new));
 
         return clusterRoleBuilder.build();
     }
 
+    @NotNull
+    private static Set<PolicyRule> getNormalizedRules(Set<PolicyRule> collectedRules) {
```
metacosm, line 137: Can you add tests for this method and maybe add more comments about what it's doing?

metacosm (changes requested):

core/deployment/src/test/java/io/quarkiverse/operatorsdk/test/OperatorSDKTest.java
```diff
@@ -108,6 +99,8 @@ public void shouldCreateRolesAndRoleBindings() throws IOException {
                             .filter(rule -> rule.getResources().equals(List.of(RBACRule.ALL)))
                             .anyMatch(rule -> rule.getVerbs().equals(List.of(UPDATE))
                                     && rule.getApiGroups().equals(List.of(RBACRule.ALL))));
+                    rules.stream().filter(rule -> rule.getApiGroups().equals(List.of(TypelessKubeResource.GROUP)))
```
metacosm, line 102: This is missing an assertion.

metacosm:
Generally speaking, I will have some improvements, especially after https://github.com/operator-framework/java-operator-sdk/pull/2515 is available in a release, as this will be more precise.

chendouble:
> Generally speaking, I will have some improvements, especially after [operator-framework/java-operator-sdk#2515](https://github.com/operator-framework/java-operator-sdk/pull/2515) is available in a release, as this will be more precise.

I agree.

metacosm:
Thank you for the improvements. However, I think that the merging of rules targeting the same apiGroup/kind should be done while we're gathering them, not after, as this would be clearer to understand, in my opinion. Would you mind modifying your PR to do that?

chendouble:
Maybe the naming is the key, not about before or after.

metacosm (approved):
Thank you!

PR #937, opened 2024-08-26
feat: improve RBAC generation for typeless dependents

metacosm:
- **refactor: merge rules as they are collected**
- **feat: throw an exception if class is not found in index**
- **fix: use plural form instead of wildcard, adapt tests**

### DIFF
--- a/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
+++ b/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
@@ -12,6 +12,7 @@ public class RBACVerbs {
     public static final String WATCH = "watch";
     public static final String DELETE = "delete";
     public static final String[] UPDATE_VERBS = new String[] { PATCH, UPDATE };
+    public static final String[] CREATE_VERBS = new String[] { CREATE, PATCH };
     public static final String[] READ_VERBS = new String[] { GET, LIST, WATCH };
     public static final String[] ALL_COMMON_VERBS;
 
--- a/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
+++ b/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
@@ -12,7 +12,6 @@ public class RBACVerbs {
     public static final String WATCH = "watch";
     public static final String DELETE = "delete";
     public static final String[] UPDATE_VERBS = new String[] { PATCH, UPDATE };
-    public static final String[] CREATE_VERBS = new String[] { CREATE, PATCH };
     public static final String[] READ_VERBS = new String[] { GET, LIST, WATCH };
     public static final String[] ALL_COMMON_VERBS;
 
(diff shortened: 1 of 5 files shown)


annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
```diff
@@ -12,6 +12,7 @@ public class RBACVerbs {
     public static final String WATCH = "watch";
     public static final String DELETE = "delete";
     public static final String[] UPDATE_VERBS = new String[] { PATCH, UPDATE };
+    public static final String[] CREATE_VERBS = new String[] { CREATE, PATCH };
```
xstefank, line 15: PATCH should create if not exist? It just seems a little off to me.

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddClusterRolesDecorator.java
```diff
@@ -57,21 +66,40 @@ public void visit(KubernetesListBuilder list) {
...
 
-    @NotNull
+    private static void collectAndMergeIfNeededRulesFrom(Collection<PolicyRule> newRules,
+            Map<String, PolicyRule> existingRules) {
+        newRules.forEach(newPolicyRule -> {
+            final var key = getKeyFor(newPolicyRule);
+            existingRules.merge(key, newPolicyRule, (existing, npr) -> {
+                Set<String> verbs1 = new TreeSet<>(existing.getVerbs());
+                verbs1.addAll(npr.getVerbs());
+                existing.setVerbs(verbs1.stream().toList());
```
xstefank, line 92: maybe just `new ArrayList<>(verbs1)` or `List.of(verbs1)` would be better so we don't need to create stream

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/AddClusterRolesDecorator.java
```diff
@@ -81,42 +109,37 @@ private static Set<PolicyRule> getClusterRolePolicyRulesFromDependentResources(Q
...
-                            @SuppressWarnings("rawtypes")
-                            GenericKubernetesDependentResource genericKubernetesResource = (GenericKubernetesDependentResource) dependentResourceClass
-                                    .getConstructor().newInstance();
-                            resourceGroup = genericKubernetesResource.getGroupVersionKind().getGroup();
-                            resourcePlural = "*";
-                        }
+                        @SuppressWarnings({ "unchecked", "rawtypes" })
+                        final var genericKubernetesResource = Utils.instantiate(
+                                (Class<? extends GenericKubernetesDependentResource>) dependentResourceClass,
+                                GenericKubernetesDependentResource.class, "AddClusterRolesDecorator");
```
xstefank, line 119: constant? Not sure how much "AddClusterRoleDecorator" could change.

annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/RBACVerbs.java
metacosm: I don't recall why we added patch there but I seem to recall it might be required when you use SSA.

metacosm: That said, you're right, PATCH should only be added if SSA is used.

xstefank (approved):

PR #939, opened 2024-08-29
fix: ignore vertx warning in reflective hierarchy during compilation

### LINKED ISSUE #936: Warning: [vertx, okhttp3, okio] Classes not in Jandex index
I see the following build warning:

maven:
```
[WARNING] [io.quarkus.deployment.steps.ReflectiveHierarchyStep] Unable to properly register the hierarchy of the following classes for reflection as they are not in the Jandex index:
	- io.vertx.core.Vertx (source: <unknown>)
	- io.vertx.core.http.HttpClient (source: <unknown>)
	- io.vertx.core.http.WebSocket (source: <unknown>)
	- io.vertx.core.net.ProxyType (source: <unknown>)
Consider adding them to the index either by creating a Jandex index for your dependency via the Maven plugin, an empty META-INF/beans.xml or quarkus.index-dependency properties.
```

gradle:
```
Unable to properly register the hierarchy of the following classes for reflection as they are not in the Jandex index:
	- io.vertx.core.Vertx (source: <unknown>)
	- io.vertx.core.http.HttpClient (source: <unknown>)
	- io.vertx.core.http.WebSocket (source: <unknown>)
	- io.vertx.core.net.ProxyType (source: <unknown>)
	- okhttp3.OkHttpClient (source: <unknown>)
	- okhttp3.OkHttpClient$Builder (source: <unknown>)
	- okhttp3.Request$Builder (source: <unknown>)
	- okhttp3.WebSocket (source: <unknown>)
	- okio.BufferedSource (source: <unknown>)
Consider adding them to the index either by creating a Jandex index for your dependency via the Maven plugin, an empty META-INF/beans.xml or quarkus.index-dependency properties.
```
I can reproduce this just by creating a [starter project](https://code.quarkus.io) that includes `quarkus-operator-sdk` and adding this file in src/main/java/MyDependentResource.java:

```
import io.fabric8.kubernetes.api.model.Service;
import io.fabric8.kubernetes.api.model.apps.Deployment;
import io.javaoperatorsdk.operator.api.reconciler.Context;
import io.javaoperatorsdk.operator.api.reconciler.EventSourceContext;
import io.javaoperatorsdk.operator.api.reconciler.ResourceDiscriminator;
import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependent;
import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResource;
import java.util.Optional;

@KubernetesDependent
public class MyDependentResource extends KubernetesDependentResource<Service, Deployment> implements ResourceDiscriminator<Service, Deployment> {
    EventSourceContext<Deployment> eventSourceContext;

    public MyDependentResource() {
        super(Service.class);
    }

    @Override
    public Optional<Service> distinguish(Class<Service> resource, Deployment primary, Context<Deployment> context) {
        return null;
    }
}
```

xstefank:
Fixes #936

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
@@ -36,6 +36,7 @@ import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.*;
 import io.quarkus.deployment.builditem.nativeimage.ForceNonWeakReflectiveClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyIgnoreWarningBuildItem;
 import io.quarkus.deployment.metrics.MetricsCapabilityBuildItem;
 import io.quarkus.gizmo.AssignableResultHandle;
 import io.quarkus.gizmo.MethodCreator;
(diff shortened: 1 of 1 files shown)


metacosm (commented):
Hmm, I'm not sure that this is the appropriate solution… 🤔 
I'd rather avoid register even more classes than we already do for reflection. Not sure why they show up now when this doesn't use the be the case before or why these HTTP client are even requiring reflective access…

xstefank:
@metacosm from what I understand, this is a static analysis during build time. We really don't know at this point if they need to be accessed reflectively.

metacosm:
> @metacosm from what I understand, this is a static analysis during build time. We really don't know at this point if they need to be accessed reflectively.

Yes, but I would like to understand why they are considered now, when they previously weren't and why we would even need to consider them, for that matter.

metacosm:
That old? That surprises me as I've never seen this issue or heard about it before! 

My guess would be that stems from the fact that which HTTP client the Fabric8 client was made configurable at some point.

Maybe @manusa has a better idea of what's going on?

manusa:
> Maybe @manusa has a better idea of what's going on?

I'm not really sure about what's going on.

My bet is that something changed in Quarkus' analysis tools and now this is being reported while it wasn't reported before.

I don't think any of those classes should be accessed reflectively (but I may be wrong).
It seems that the kubernetes-client-extension processing might be very greedy and is registering classes that reference the ones reported. Other than that, I cannot think of any other reason as to why this is happening.

xstefank:
Got it - https://github.com/quarkiverse/quarkus-operator-sdk/commit/b8d79851a2bf2ea8a790386d67424ff7a07a8912.

metacosm:
> Got it - [b8d7985](https://github.com/quarkiverse/quarkus-operator-sdk/commit/b8d79851a2bf2ea8a790386d67424ff7a07a8912).

Ah, so you're saying it's my fault! 😀 

> 
> So it was always there. This change was just to move from reflective class registration to reflective hierarchy registration so it shows the warning messages.

Several todos in that code, I think it's time to clean that up! Thank you for identifying the cause.

xstefank:
@metacosm I think we can merge this for the time being and we can take a look at the better solution later.

metacosm:
> @metacosm I think we can merge this for the time being and we can take a look at the better solution later.

Can you create an issue to clean this up at a later point, please, then?

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
```diff
@@ -261,6 +262,27 @@ void initializeRuntimeNamespacesFromBuildTimeValues(
 
     }
 
+    /**
+     * Ignore warnings related to non-indexed classes in the reflective hierarchy. At this point, we cannot know
+     * if they are actually needed for native compilation.
+     */
```
metacosm, line 268: Maybe add a comment that this might get removed once the underlying issue is addressed?

metacosm (approved):

core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/OperatorSDKProcessor.java
xstefank: Just for reference - https://github.com/quarkiverse/quarkus-operator-sdk/issues/941

xstefank:
Just for reference https://github.com/quarkiverse/quarkus-operator-sdk/issues/941

PR #943, opened 2024-09-02
chore: update version to 6.8.0-SNAPSHOT

metacosm:


xstefank (commented):
approving, just to verify I looked

PR #944, opened 2024-09-02
fix: add ConfigurationService implementation to support build-time resolution

metacosm:
- **feat: add build-time ConfigurationService implementation**
- **fix: only add PATCH verb if resource is creatable**


xstefank (commented):
LGTM

PR #947, opened 2024-09-05
feat: provide per-bundle configuration via properties

### LINKED ISSUE #903: Allow Bundle CSV annotations to be given via configuration
We can currently add annotations to the CSV using `@CSVMetadata`, but some annotations are dynamic and would be better given via configs.

For example:
- `olm.skipRange`
- `createdAt`
- `containerImage` (bundle generator could probably optionally figure this one out automatically)

Maybe a general purpose solution could be to allow configuration of arbitrary annotations via config like:
```
quarkus.operator-sdk.bundle.annotations."my-annotation"=annotation-value
```

metacosm:
Fixes #903


MikeEdgar (approved):

xstefank (commented):
LGTM

PR #948, opened 2024-09-05
feat: use minikube instead of KinD

metacosm:


xstefank (approved):
Why this change @metacosm?

I couldn't run kind on my machine so I would happy with this, just wondering why :).

metacosm:
The idea is to be able to test against multiple Kubernetes versions as we do in JOSDK. This currently doesn't work because we need to find a way to expose the minikube internal registry similarly to what is done with KinD.

xstefank:
I think we can drop this for now.

metacosm:
I would still like to support testing against multiple kube versions at some point but let's revisit this then.

PR #952, opened 2024-09-10
feat: use service account determined by Quarkus

metacosm:

PR #953, opened 2024-09-10
fix: use service account name instead of deployment name

### LINKED ISSUE #932: Fine-tuning service account namespace
We can currently give the service account's namespace using `quarkus.kubernetes.namespace`, but this also results in the namespace being added everywhere to the generated Kubernetes resources. It would be useful to be able to also use `quarkus.kubernetes.rbac.service-accounts.my-service-account.namespace=my_namespace` as the primary lookup, then fall-back to `quarkus.kubernetes.namespace`.

Aside - while looking at `AddRoleBindingsDecorator`, I noticed that the service account name is based on the deployment name. If the deployment has its own `serviceAccountName` already, it is ignored for the role binding, but still used in a generated CSV for the deployment.

metacosm:
Fixes #932

### DIFF
--- a/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/ConfiguredServiceAccountNameShouldBeUsedTest.java
+++ b/bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/ConfiguredServiceAccountNameShouldBeUsedTest.java
@@ -1,0 +1,44 @@
+package io.quarkiverse.operatorsdk.bundle;
+
+import static io.quarkiverse.operatorsdk.bundle.Utils.BUNDLE;
+import static io.quarkiverse.operatorsdk.bundle.Utils.getCSVFor;
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.io.IOException;
+import java.nio.file.Files;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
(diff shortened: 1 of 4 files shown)


bundle-generator/deployment/src/test/java/io/quarkiverse/operatorsdk/bundle/ConfiguredServiceAccountNameShouldBeUsedTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.quarkiverse.operatorsdk.bundle.sources.First;
+import io.quarkiverse.operatorsdk.bundle.sources.ReconcilerWithNoCsvMetadata;
+import io.quarkus.test.ProdBuildResults;
+import io.quarkus.test.ProdModeTestResults;
+import io.quarkus.test.QuarkusProdModeTest;
+
+public class ConfiguredServiceAccountNameShouldBeUsedTest {
+
+    public static final String APPLICATION_NAME = "configured-service-account-name";
```
xstefank, line 21: @metacosm isn't this conflicting your other draft PR?

metacosm: Yes, it is, I created the other one to show current progress and discuss things with @iocanel. This one will get a workaround with the current code to provide a fix but, ultimately, the proper way to handle things should be what has been started in the other PR.

metacosm:
Closed in favor of #952.

PR #976, opened 2024-10-16
Split and Reuse the Release workflow

gastaldi:
- Needed for https://github.com/quarkiverse/quarkiverse-devops/issues/275. If you have customized the release process for your case, make sure they are properly adjusted, as this PR is automatically generated


.github/workflows/release-prepare.yml
```diff
@@ -0,0 +1,28 @@
+name: Quarkiverse Perform Release
```
metacosm, line 1: Name of the file doesn't match the name of the workflow

.github/workflows/release-perform.yml
```diff
@@ -0,0 +1,18 @@
+name: Quarkiverse Prepare Release
```
metacosm, line 1: Name of the file doesn't match workflow name

metacosm (changes requested):

metacosm:
Our release process is customized so I cannot merge these changes as-is and will probably require to do it manually, anyway.

PR #977, opened 2024-10-16
chore(ci): update release process according to new Quarkiverse process

metacosm:


.github/workflows/release-perform.yml
```diff
@@ -0,0 +1,105 @@
...
+        id: metadata
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          metadata-file-path: '.github/project.yml'
+          local-file: true
+
+      - name: Gather versions
+        id: versions
+        run: |
+          echo "quarkus_version=$(./mvnw help:evaluate -Dexpression=quarkus.version -q -DforceStdout)" >> $GITHUB_OUTPUT
```
gastaldi, line 54: You can save some time by specifying `${{github.event.inputs.tag || github.ref_name}}` directly here

gastaldi (approved):
LGTM, added a minor suggestion, which you can ignore if it doesn't make sense

.github/workflows/release-perform.yml
```diff
@@ -0,0 +1,105 @@
...
+          fi
+
+      - name: Wait for QOSDK BOM to be published
+        uses: quarkusio/monitor-artifact-publication-action@main
+        id: wait-for-bom
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          group-id: io.quarkiverse.operatorsdk
+          artifact-id: quarkus-operator-sdk-bom
+          version: ${{steps.metadata.outputs.current-version}}
```
gastaldi, line 89: This should also work: 

```suggestion
          version: ${{github.event.inputs.tag || github.ref_name}}
```

.github/workflows/release-perform.yml
metacosm: I don't think this would work here as this specific line is supposed to retrieve the Quarkus version defined in the POM, not the QOSDK version.

gastaldi:
Don't forget to delete https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/.github/workflows/release.yml in this PR too

.github/workflows/release-perform.yml
metacosm: I was suspecting as much but wasn't 100% sure.

.github/workflows/release-perform.yml
gastaldi: Ah you're right, I missed that. Good call

.github/workflows/release.yml
```diff
@@ -1,125 +0,0 @@
...
-              echo "PR targets main so assume quarkus-platform main also needs to be updated"
-              echo "need_main_pr=true" >> $GITHUB_OUTPUT
-            else
-              echo "No need for PR on quarkus-platform main"
-              echo "need_main_pr=false" >> $GITHUB_OUTPUT
-            fi
-          fi
-          
-          git checkout -b release
-          ./mvnw -B release:prepare -Prelease -Darguments="-DperformRelease -Dno-samples -DskipTests" -DreleaseVersion=${{steps.metadata.outputs.current-version}} -DdevelopmentVersion=${{steps.metadata.outputs.next-version}}
```
gastaldi, line 90: I see you're passing a `-Dno-samples` here, which is not done in the reusable workflow. Can you define it in a Maven profile?

metacosm: Though that raises an interesting question that I hadn't paid attention to, indeed. How can I not release my samples with this new process?

gastaldi: https://github.com/quarkiverse/quarkus-ironjacamar/blob/main/pom.xml#L70-L73

metacosm: Hmm, though that means I need another profile to deactivate based on the property…

gastaldi: According to https://maven.apache.org/guides/introduction/introduction-to-profiles.html#multiple-conditions, you can specify more than one activation. Plus you can also specify `!true` to match "any value but true" in the `no-*` properties you already have: 
 
```xml
    <profile>
      <id>no-docs</id>
      <activation>
        <property>
          <name>no-docs</name>
          <value>!true</value>     
        </property>
        <property>
          <name>performRelease</name>
          <value>!true</value>
        </property>
      </activation>
      <modules>
        <module>docs</module>
      </modules>
    </profile>
```

metacosm: IDEA doesn't agree that you can specify multiple properties. Also, the semantics is supposed to be AND, which isn't what I would like (OR semantics).

xstefank (approved):
If the new profile `no-samples-in-release` solves the issue, LGTM.

@gastaldi please can you check again

gastaldi (approved):

PR #982, opened 2024-10-29
fix: issue a warning if CRD couldn't be applied

metacosm:
- **fix: unwrap optional**
- **fix: issue a warning if CRD couldn't be applied**


.github/workflows/pre-release.yml
```diff
@@ -5,21 +5,12 @@ on:
...
-        if: contains(steps.metadata.outputs.current-version, 'SNAPSHOT')
-        run: |
-          echo '::error::Cannot release a SNAPSHOT version.'
-          exit 1
\ No newline at end of file
+jobs:
+  pre-release:
+    name: Pre-Release
+    uses: quarkiverse/.github/.github/workflows/pre-release.yml@main
+    secrets: inherit
```
xstefank, line 16: Shouldn't this go in a separate PR?

xstefank (changes requested):
@metacosm I think you mixed CI changes in this PR and that should go in a different commit message.

.github/workflows/pre-release.yml
xstefank: my bad, I didn't see that this is actually multiple commits.

xstefank (approved):

metacosm:
I'm actually not sure of what's going on there, I think I will re-open the PR in a cleaner way as I originally had it targeting the wrong branch.

PR #990, opened 2024-10-31
chore: align generated docs file in git

xstefank:
@metacosm I'm not sure if this is only on my PC, but these 4 files change with every build so I would like to add them to git.


metacosm (approved):
Yes, these files got changed with the Quarkus 3.16 update.

PR #992, opened 2024-11-07
feat: do not limit external CRDs to bundle generation

metacosm:
- **refactor: move external CRD configuration to main CRD configuration**
- **refactor: load external CRDs from CRD generation step**
- **refactor: simplify add method**
- **feat: only process CRs that are not excluded**


andreaTP (approved):
LGTM, eager to try out all those improvements as soon as I have an occasion!
Thanks!

xstefank (commented):
LGTM

PR #1001, opened 2024-11-14
chore(deps): update to Quarkus 3.17.0.CR1

metacosm:

### DIFF
--- a/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
+++ b/bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
@@ -121,7 +121,7 @@ public class BundleProcessor {
 
         // note that version, replaces, etc. should probably be settable at the reconciler level
         // use version specified in bundle configuration, if not use the one extracted from the project, if available
-        final var version = kubernetesConfig.getVersion().orElse(appConfiguration.getVersion());
+        final var version = kubernetesConfig.version().orElse(appConfiguration.getVersion());
         final var defaultVersion = bundleConfiguration.version()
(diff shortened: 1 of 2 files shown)


bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -121,7 +121,7 @@ CSVMetadataBuildItem gatherCSVMetadata(KubernetesConfig kubernetesConfig,
 
         // note that version, replaces, etc. should probably be settable at the reconciler level
         // use version specified in bundle configuration, if not use the one extracted from the project, if available
-        final var version = kubernetesConfig.getVersion().orElse(appConfiguration.getVersion());
+        final var version = kubernetesConfig.version().orElse(appConfiguration.getVersion());
```
xstefank, line 124: Not sure if we should push different changes in the update commits.

pom.xml
```diff
@@ -79,6 +79,7 @@
       <properties>
         <skipTests>true</skipTests>
         <skipITs>true</skipITs>
+        <skipDocs>true</skipDocs>
```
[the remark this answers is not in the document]
metacosm: Weird…

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
metacosm: This is a required change. The code won't compile without this change. This is the reason why the ecosystem checks have been failing for so long as well.

xstefank: OK, thanks for the explanation.

PR #1007, opened 2024-11-22
feat: make it possible to revert to using the v1 CRD generator

metacosm:
Note that this should only be used if a blocking issue is found with the
v2 generator because the v1 version won't be supported moving forward.


metacosm:
cc @jcechace

xstefank (commented):
LGTM

PR #1010, opened 2024-11-25
doc: clarify that CRDs are never applied in prod mode

metacosm:

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/CRDConfiguration.java
+++ b/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/CRDConfiguration.java
@@ -32,7 +32,10 @@ public interface CRDConfiguration {
 
     /**
      * Whether the extension should automatically apply updated CRDs when they change.
-     * When running on DEV mode, the CRD changes will always be applied automatically.
+     * <p>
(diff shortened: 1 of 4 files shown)


core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/CRDConfiguration.java
```diff
@@ -32,7 +32,10 @@ public interface CRDConfiguration {
 
     /**
      * Whether the extension should automatically apply updated CRDs when they change.
-     * When running on DEV mode, the CRD changes will always be applied automatically.
+     * <p>
+     * <strong>NOTE that this option only when *not* in Dev mode as applying the CRD to a production cluster could be
```
xstefank, line 36: missing verb?

xstefank: I'm not sure what this is saying. It applies only in dev and test, correct?

metacosm: D'oh! 😓 Opened a new PR to fix this. Thank you!

PR #1012, opened 2024-11-28
fix: inject client instead of creating it manually

metacosm:
- **fix: inject client instead of creating it manually**
- **chore: update generated docs**


xstefank (commented):
LGTM

PR #1026, opened 2025-01-20
chore: move Quarkus platform PR info to the project.yml file

### LINKED ISSUE #996: Improve Quarkiverse platform update process on release
Currently, the process is not covering all cases, which leads to missed platform updates opportunities and frustration because platform PRs then need to be opened manually.

We should extend the `project.yml` metadata so that it also includes which platform branches are targeted by a given release as this would make things more explicit and allow branches to be "locked" to a given Quarkus major (e.g. LTS version).

xstefank:
Fixes #996

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -2,4 +2,6 @@ name: Java Operator SDK Extension
 release:
   current-version: 7.0.0.CR1
   next-version: 7.0.0-SNAPSHOT
-
+  quarkus-platform-branches:
+    - "main"
+    - "3.17"
--- a/.github/workflows/release-quarkus-platform-pr.yml
+++ b/.github/workflows/release-quarkus-platform-pr.yml
@@ -6,10 +6,9 @@ on:
       qosdk-version:
         type: string
         required: true
-      quarkus-platform-branch:
+      quarkus-platform-branches:
         type: string
-        required: false
-        default: 'main'
+        required: true
     secrets:
       qosdk-bot-token:
         required: true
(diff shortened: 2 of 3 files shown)


xstefank:
@metacosm design is open for discussion. Metadata extension only takes the leaf name of the yaml, so any structure in `project.yml` is ignored.

metacosm (changes requested):
There needs to be potentially several target branches as a platform PR might need to be opened against `main` but also the current stable version (for example, right now, `main` and `3.17` potentially)

.github/project.yml
```diff
@@ -2,4 +2,6 @@ name: Java Operator SDK Extension
 release:
   current-version: 7.0.0.CR1
   next-version: 7.0.0-SNAPSHOT
-
+  quarkus-platform:
+    target-branch: main
+    needs-main-pr: false
```
metacosm, line 7: I guess this shouldn't be needed anymore if we specify branches to target explicitly

metacosm:
> @metacosm design is open for discussion. Metadata extension only takes the leaf name of the yaml, so any structure in `project.yml` is ignored.

Ah, too bad. In that case, the leaf should probably contain `quarkus-platform` in some form in its name to avoid potential collisions?

.github/project.yml
```diff
@@ -2,4 +2,6 @@ name: Java Operator SDK Extension
 release:
   current-version: 7.0.0.CR1
   next-version: 7.0.0-SNAPSHOT
-
+  quarkus-platform-branches:
+    - "main"
+    - "3.17"
```
xstefank, line 7: @metacosm these don't need to be strings, but otherwise we will get warnings that we mix strings and numbers. I mean, it can also be without quotes like this if preferred:

```yaml
  quarkus-platform-branches:
    - main
    - 3.17
```

metacosm: Probably better with quotes.

metacosm (approved):

PR #1044, opened 2025-02-09
chore: update to JOSDK 5.0.1

metacosm:
- **chore: update to JOSDK 5.0.1**
- **chore: release 7.0.1**


xstefank (commented):
hm, this is just my opinion but I think we shouldn't mix QOSDK updates with project.yml updates.

metacosm:
The PR was merged with a rebase so it will appear as different commit in the history but, yes, it's probably better to do these changes in separate PRs.

PR #1069, opened 2025-03-24
feat: apply external CRDs during test / dev mode

metacosm:
- **refactor: use CRD spec version from the info**
- **feat: apply external CRDs during test / dev mode**

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -18,10 +18,6 @@
       <groupId>io.quarkiverse.operatorsdk</groupId>
       <artifactId>quarkus-operator-sdk</artifactId>
     </dependency>
-   <!-- <dependency>
-      <groupId>io.quarkiverse.operatorsdk</groupId>
-      <artifactId>quarkus-operator-sdk-bundle-generator</artifactId>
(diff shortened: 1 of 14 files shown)


integration-tests/pom.xml
```diff
@@ -18,10 +18,10 @@
       <groupId>io.quarkiverse.operatorsdk</groupId>
       <artifactId>quarkus-operator-sdk</artifactId>
     </dependency>
-    <dependency>
+   <!-- <dependency>
       <groupId>io.quarkiverse.operatorsdk</groupId>
       <artifactId>quarkus-operator-sdk-bundle-generator</artifactId>
-    </dependency>
+    </dependency>-->
```
xstefank, line 24: intentional or forgotten test?

xstefank (approved):

integration-tests/pom.xml
metacosm: Good point! Need to double check but if it works nicely without, might as well keep it removed since this part shouldn't be checking the bundle generation and therefore could be more efficient without that useless processing.

PR #1090, opened 2025-05-07
feat: allow reconcilers to override their configuration at runtime

metacosm:
If a reconciler implements `ConfigurableReconciler`, then QOSDK will
derive a new configuration for the reconciler by calling
`updateConfigurationFrom` before registering the reconciler with this
new configuration.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/OperatorProducer.java
+++ b/core/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/OperatorProducer.java
@@ -12,7 +12,10 @@ import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.Operator;
+import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
+import io.javaoperatorsdk.operator.api.config.ControllerConfigurationOverrider;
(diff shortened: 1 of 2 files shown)


xstefank (approved):
I think this is better. But wasn't there also a test?

metacosm:
> I think this is better. But wasn't there also a test?

There isn't at the moment, but I need to add some, indeed.
@goldmann would this solution work for you?

goldmann:
Yes, yep! Will work just fine!

goldmann:
I'm a little bit puzzled on why this was closed. Isn't this the extension repo where this should be living? Or maybe I'm understanding it wrongly? What is the relation of https://github.com/operator-framework/java-operator-sdk/pull/2784 to this PR?

Thanks and sorry!

metacosm:
I'm puzzled as well! 😅 
I guess I was tired and closed the wrong PR. 🤦🏼 
Thank you!

metacosm:
> I'm a little bit puzzled on why this was closed. Isn't this the extension repo where this should be living? Or maybe I'm understanding it wrongly? What is the relation of [operator-framework/java-operator-sdk#2784](https://github.com/operator-framework/java-operator-sdk/pull/2784) to this PR?

That PR is a change I thought would be needed but isn't after all. This is a tangential issue that is also kind of orthogonal (if that makes any sense!) 😁 
Strictly speaking, one is not needed for the other.

Issue #1088, opened 2025-05-03
Adding labels to ClusterServiceVersion about supported architectures


ahus1:
To ship an Operator for ARM, [the docs state that I need to add labels](https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/#multiple-architectures) to the CSV.

I looked at [CSVMetadata](https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/annotations/src/main/java/io/quarkiverse/operatorsdk/annotations/CSVMetadata.java) and didn't see a way to add labels here. 

Am I missing something here, or should I ask for a RFE to have those labels added? 

Thanks!

PR #1094, opened 2025-05-20
fix: Allow to add labels in CSVMetadata

xstefank:
If accepted, should be ported to main for #1088.


bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
```diff
@@ -495,6 +488,19 @@ private CSVMetadataHolder createMetadataHolder(AnnotationInstance csvMetadata, C
                 origin);
     }
 
+    private Map<String, String> extractNameValueMap(AnnotationValue labelsField) {
```
metacosm, line 491: Nitpicking: can you rename the parameter and variables more generically, please?

metacosm (changes requested):

bundle-generator/deployment/src/main/java/io/quarkiverse/operatorsdk/bundle/deployment/BundleProcessor.java
xstefank: omg, thx

metacosm (approved):

PR #1119, opened 2025-06-27
feat: add a new helm module that builds on top of quarkus-helm extension

### LINKED ISSUE #919: Separate Helm generator from core?
Helm generation should perhaps be dealt with in a way similar to the OLM generation, via a separate extension module.

xstefank:
Resolves #919

Still needs a lot of cleaning and more tests. Opening the PR just to double check the approach.


xstefank:
Closing as we decided to go with a different approach.

PR #1135, opened 2025-08-05
fix: remove test-kubernetes-client from the joke sample to use provided Dev Services

xstefank:
It seems that before https://github.com/quarkusio/quarkus/pull/48977 the `quarkus-test-kubernetes-client` was not correctly recognized and the Kubernetes Client Dev Services were able to start the container. But after this change the extension presence is correctly detected and thus the containers are not started.

### DIFF
--- a/samples/joke/pom.xml
+++ b/samples/joke/pom.xml
@@ -61,11 +61,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-test-kubernetes-client</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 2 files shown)


metacosm:
@xstefank any idea?

holly-cummins:
Or ... thinking about it, and your changeset, I think I maybe did change the behaviour slightly, from "are any of the test client classes _used_" to "are any of the test client classes on the classpath?"

xstefank:
> Or ... thinking about it, and your changeset, @xstefank, I think I maybe did change the behaviour slightly, from "are any of the test client classes _used_" to "are any of the test client classes on the classpath?"
> 
> The (docs)[https://quarkus.io/guides/kubernetes-dev-services] do say "if you include the quarkus-test-kubernetes-client dependency," so if so, the good news is that the new behaviour is closer to what's documented. But the bad news is it's different.

It is definitely in the classloader instance that is used to call `Package.getPackages()`. The searched package just isn't there, no matter how I try to include the quarkus-test-kubernetes-client extension. I think your comment https://github.com/quarkusio/quarkus/pull/48977/files#diff-7078a650839666ccd342125978a7ba1fd911a01acd4d4c7862de6a8b9dae4cf1R16 might be the cause of this.

PR #1139, opened 2025-08-13
feat: integrate with quarkus-helm in a conditional way

### LINKED ISSUE #919: Separate Helm generator from core?

xstefank:
Resolved #919


samples/exposedapp/src/test/java/io/halkyon/HelmDeploymentE2EIT.java
```diff
@@ -37,21 +37,21 @@ class HelmDeploymentE2EIT {
     final static Logger log = Logger.getLogger(HelmDeploymentE2EIT.class);
 
     static class RunOnlyIfHelmIsAvailableCondition implements ExecutionCondition {
-        private static Boolean helmIsAvailable;
+        private static boolean helmIsAvailable = false;
```
metacosm, line 40: I'd rather disable the test explicitly using `@Disabled` but that's fine for now.

PR #1153, opened 2025-08-30
chore(deps): clean module build ordering

metacosm:


samples/pingpong/pom.xml
```diff
@@ -61,6 +55,11 @@
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
xstefank, line 58: Don't understand what this is needed for.

metacosm: Because if you remove it, the tests won't build 😅

xstefank (approved):

PR #1172, opened 2025-09-29
chore: release 7.2.1

xstefank:


.github/project.yml
```diff
@@ -1,7 +1,6 @@
 name: Java Operator SDK Extension
 release:
-  current-version: 7.2.0
-  next-version: 7.2.1-SNAPSHOT
+  current-version: 7.2.1
+  next-version: 7.2.2-SNAPSHOT
   quarkus-platform-branches:
```
xstefank, line 5: @metacosm double check this please if this is correct.

metacosm: This looks OK, however, this doesn't downgrade to Quarkus 3.27 as is implied by the PR title: if this PR is merged, it will release 7.2.1 and update Quarkus platform for 3.27 but won't downgrade to Quarkus 3.27.0 (which has previously been done).

PR #1209, opened 2025-11-24
fix: Helm doesn't correctly propagate role bindings

xstefank:
This is a test fix for the issue reported in the JOSDK discord.

@jcechace can you please check if this fixes your issue?

I wonder why the tests didn't catch this so I will need to investigate a little further.


xstefank:
@jcechace because I think that the `--set 'app.envs.QUARKUS_OPERATOR_SDK_CONTROLLERS_EXPOSEDAPP_NAMESPACES=JOSDK_WATCH_CURRENT'` works as expected.

PR #1211, opened 2025-11-26
feat: initial support for JOSDK 5.2.0

metacosm:


samples/mysql-schema/src/main/java/io/quarkiverse/operatorsdk/samples/mysqlschema/schema/Schema.java
```diff
@@ -40,4 +42,9 @@ public int hashCode() {
     public String toString() {
         return "Schema{" + "name='" + name + '\'' + ", characterSet='" + characterSet + "'}";
     }
+
+    @Override
+    public String resourceId() {
+        return getName();
```
xstefank: this could just directly return field but no biggie

xstefank: Then maybe add a comment with this? I think it might be important for someone checking this sample.

PR #1214, opened 2025-12-01
fix: Allow Helm dynamic overrides

xstefank:


xstefank:
@jcechace can you try this one please?

jcechace:
@xstefank  will give it a try this weekend. Sorry for the delay, it was a bit busier week than anticipated. I haven't forgotten though!

PR #1235, opened 2025-12-18
feat: fail at build time if the external resource doesn't implement ResourceIDProvider

### LINKED ISSUE #1212: Detect missing implementations of `ResourceIDProvider` at build time?

xstefank:
fixes #1212


core/deployment/src/main/java/io/quarkiverse/operatorsdk/deployment/QuarkusControllerConfigurationBuildStep.java
```diff
@@ -520,6 +527,31 @@ private static DependentResourceSpecMetadata createDependentResourceSpec(
         return spec;
     }
 
+    private static void failOnMissingResourceIDProvider(String dependentTypeName, String resourceTypeName, IndexView index) {
+        if (!index.getAllKnownImplementations(DotName.createSimple(ResourceIDMapper.class)).isEmpty()) {
```
xstefank: @metacosm do we know if the `ResourceIDMapper` is in use in some of the event sources?

xstefank:
Just to paste my investigation here too:

I don't think it's possible to find if the `ResourceIDMapper` was overriden or not at build time. It is passed in 3 cases:

`ExternalResourceCachingEventSource` -> in constructor -- not possible at build time
`ExternalBulkDependentResource` -> default interface method -- this we could detect at build time
`AbstractExternalDependentResource` -> public setter, so I don't know if Jandex maybe can find usages of the setter at build time, but if not, also not possible to do at build time

PR #1242, opened 2026-01-16
feat: move baseline to Java 21

metacosm:


xstefank (changes requested):
I don't think we should move our baseline over the Quarkus's one and Quarkus is still baseline JDK 17.

PR #1302, opened 2026-03-14
chore: restore evergreen versions for next branch

metacosm:


xstefank (approved):
Why still draft?

xstefank:
Actually, I missed this is aiming for main. So probably ok to close since we now have CI testing for this?

metacosm:
I guess we could close it, I keep it around because we sometimes require code changes to support the next JOSDK version.

xstefank:
Wouldn't it be better then to have a ci job to test any custom branch with this?

metacosm:
> Wouldn't it be better then to have a ci job to test any custom branch with this?

We would still need to create the CI job for it, though but I agree that we don't need to keep a long-running branch if we don't need any code changes.

PR #1348, opened 2026-04-20
refactor: leverage already needed buildnum plugin instead of git-commit

metacosm:
- **refactor: leverage already needed buildnum plugin instead of git-commit**
- **fix: remove unneeded configuration and plugins**


xstefank (commented):
Where is the buildnumber plugin defined?

xstefank (approved):
Answered offline, it is in quarkiverse parent.

PR #1351, opened 2026-04-23
fix: split rest-assured verifications to individual method calls

xstefank:
I believe this should fix groovy errors we see in the platform testing.


metacosm:
Any idea why or what changed that would make it fail?

metacosm:
For reference, we believe the issue should be addressed at the platform CI level, not for each individual projects so we're closing this.

PR #1355, opened 2026-04-29
fix: separate fabric8 client testing into separate workflow

xstefank:
-

### DIFF
--- a/.github/workflows/fkc-snapshot.yml
+++ b/.github/workflows/fkc-snapshot.yml
@@ -1,0 +1,14 @@
+name: Build main branch against latest Fabric8 client snapshot
+
+on:
+  schedule:
+    - cron: '0 23 * * 1'
+  workflow_dispatch:
+
+jobs:
+  build-with-fkc-snapshot:
+    uses: ./.github/workflows/build-for-quarkus-version.yml
+    with:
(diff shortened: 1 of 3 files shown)


.github/workflows/fkc-snapshot.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  schedule:
+    - cron: '0 23 * * 1'
+  workflow_dispatch:
+
+jobs:
+  build-with-fkc-snapshot:
+    uses: ./.github/workflows/build-for-quarkus-version.yml
+    with:
+      fkc-pr: main
```
metacosm, line 12: As discussed this morning, this isn't going to be enough. When the build asks for a non-Quarkus-specified client version, that version should be passed as a build parameter to activate the appropriate profile. I think the easiest version is to have a step that appends the appropriate incantation (`-Dfkc.version=<FKC version to use>`) to the `mvnArgs` parameter.

metacosm (changes requested):

metacosm:
Also, please fix the commit message ;)

metacosm (approved):

xstefank:
sry @metacosm, misclicked

PR #1367, opened 2026-05-18
[next] fix: explicitly define maven-install-plugin version

xstefank:


metacosm (approved):

gastaldi:
We should add this to quarkiverse-parent

PR #1371, opened 2026-05-23
docs: include attributes.adoc in deploy-with-olm page

gastaldi:
## Summary
- The page uses {project-version} but never included the attributes file where it's defined. 
- Add the include directive, matching the pattern used by index.adoc
- Fixes 1 warning during the quarkiverse-docs Antora build

## Test plan
- [ ] Verify the Maven dependency snippet renders the correct version in the documentation


metacosm:
Thank you!

PR #1386, opened 2026-06-29
fix(tmp): replace removed annotation

metacosm:
Note that this is, maybe, a temporary fix since children commands are
not currently properly detected with this fix.
It has, however, the advantage of being backwards-compatible for the time being.


xstefank (approved):
OK, we need to follow up with solution for the CLI.

PR #1397, opened 2026-07-17
chore(deps): bump quarkus to 3.38.0.CR1

xstefank:


xstefank:
this passed with 3.38.0.CR1, the failing is with 3.37.3 that doesn't have the native fix backported.

PR #1409, opened 2026-08-13
fix: uninstall OLM samples between installs to avoid resolver conflicts

xstefank:
Should fix https://github.com/quarkiverse/quarkus-operator-sdk/actions/runs/31613652905/job/94170976843

### DIFF
--- a/.github/scripts/installOperatorUsingOlm.sh
+++ b/.github/scripts/installOperatorUsingOlm.sh
@@ -66,8 +66,16 @@ spec:
   sourceNamespace: $K8S_NAMESPACE
 EOF
 
+# Discover the CSV name via the subscription: it's not necessarily "$bundle_name.v<version>"
+# (e.g. Ping Pong's CSV is "quarkus-operator-sdk-samples-pingpong.v<version>"), but OLM always
+# names CSVs "<pkg>.v<version>", so wait for installedCSV to contain ".v" once it's populated.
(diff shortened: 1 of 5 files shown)


cmoulliard (approved):
I don't know very well the project quarkus-operator-sdk but the scripts seems good

metacosm:
Could you provide more details as to why this is needed now when it used to work fine? Not saying that this change isn't needed, just trying to understand why we didn't need to do this before. What problem does it solve and why does that problem exist?

xstefank:
OLM's resolver solves Subscription in a namespace whenever a new one is created. Because Joke already succeeded, its CSV was still visible as both an existing installation and also the catalog candidate for the same CRD. The resolver saw two conflicting providers and rejected the whole namespace — a race condition that we simply didn't hit before.

Cleanup is, however, I think, a correct way of handling test cases like this.

metacosm:
> OLM's resolver solves Subscription in a namespace whenever a new one is created. Because Joke already succeeded, its CSV was still visible as both an existing installation and also the catalog candidate for the same CRD. The resolver saw two conflicting providers and rejected the whole namespace — a race condition that we simply didn't hit before.
> 
> Cleanup is, however, I think, a correct way of handling test cases like this.

Thanks for the explanation. However, shouldn't we ensure that the clean-up also occurs when errors occurs? It seems like the scripts can exit before the clean-up script gets called in some error situations.

metacosm (approved):
Look re

